PR #29, opened 2021-11-07
Modified probe interval and timeout fields to time.Duration parseable strings

### LINKED ISSUE #3: Support specifying prober interval / timeouts as durations
Copied from https://github.com/google/cloudprober/issues/611, originally filed by @vadimberezniker.

It would be nice if the prober config supported specifying durations as user-friendly strings.
i.e.
```
interval: "10m" 
timeout: "15s"
```
instead of
```
interval_msec: 600000  # 10m
timeout_msec: 15000   # 15s
```

+1ed by @benner @manugarg

nvidyala:
Motivation:

Prober configuration currently only supports specifying timeout and interval periods as int32 fields representing millisecond values. This PR aims to make this configuration more user-friendly by modifying these fields to time.Duration interpretable fields.

Modifications:

- Modify type and default values within prober config proto
- Adds parsing logic when constructing an `Options` instance

Result:
- Timeout and interval periods can now be specified as more user-friendly strings
- Closes #3


nvidyala:
@manugarg I wasn't sure if the generated protobuf file needs to be committed as well. Does the build pipeline handle the creation of these files using the gen_pb_go.sh?

manugarg:
Thanks for the PR @nvidyala.

We cannot replace the existing fields as that will break all the existing configs. What we want to do here is the following:

* Add new string fields: interval, and timeout.
* While parsing config files, return an error if both {interval,timeout}_msec and {interval,timeout} are defined.

Does that make sense?

>  I wasn't sure if the generated protobuf file needs to be committed as well. Does the build pipeline handle the creation of these files using the gen_pb_go.sh?

Yes please, generated protobuf need to submitted too.

nvidyala:
@manugarg That makes sense, my bad about not analyzing the impact of this contract modification to the config file. I'll make the necessary changes.

nvidyala:
@manugarg I took a stab at fixing the PR as per your comments, please have a look and let me know if this is more aligned to what you were envisioning.

probes/options/options.go
```diff
@@ -113,14 +115,43 @@ func getSourceIPFromConfig(p *configpb.ProbeDef, l *logger.Logger) (net.IP, erro
 // BuildProbeOptions builds probe's options using the provided config and some
 // global params.
 func BuildProbeOptions(p *configpb.ProbeDef, ldLister endpoint.Lister, globalTargetsOpts *targetspb.GlobalTargetsOptions, l *logger.Logger) (*Options, error) {
+	var err error
+	var intervalDuration time.Duration
+	var timeoutDuration time.Duration
+
+	if p.GetIntervalMsec() != 0 && p.GetInterval() != "" {
+		return nil, fmt.Errorf("both interval: (%s) and interval_msec: (%d) are specified", p.GetInterval(), p.GetIntervalMsec())
```
manugarg, line 123: I'd remove the colons after interval and interval_msec:

both interval (%s) and ...

probes/options/options.go
```diff
@@ -113,14 +115,43 @@ func getSourceIPFromConfig(p *configpb.ProbeDef, l *logger.Logger) (net.IP, erro
...
+	} else if p.GetIntervalMsec() == 0 && p.GetInterval() == "" {
+		intervalDuration = defaultIntervalPeriod
```
manugarg, line 125: How about you add this to the start of the function:

```
intervalDuration, timeoutDuration := defaultIntervalPeriod, defaultTimeoutPeriod
var err error 

if p.....
```

probes/options/options.go
```diff
@@ -113,14 +115,43 @@ func getSourceIPFromConfig(p *configpb.ProbeDef, l *logger.Logger) (net.IP, erro
...
+	} else {
+		intervalDuration, err = time.ParseDuration(p.GetInterval())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse interval (%s): %v", p.GetInterval(), err)
+		}
+	}
+
+	if p.GetTimeoutMsec() != 0 && p.GetTimeout() != "" {
+		return nil, fmt.Errorf("both timeout: (%s) and timeout_msec: (%d) are specified", p.GetTimeout(), p.GetTimeoutMsec())
+	} else if p.GetTimeoutMsec() == 0 && p.GetTimeout() == "" {
```
manugarg, line 137: Same suggestion as above.

probes/proto/config.proto
```diff
@@ -131,6 +131,12 @@ message ProbeDef {
   //   }
   repeated AdditionalLabel additional_label = 14;
```
manugarg, line 132: I'd keep these field definitions close to their "msec" counterparts, even if their proto field numbers are far apart:

```
// Interval between probes in milliseconds "or" string form (e.g. 5s, 1m, etc). You can
// either specify interval_msec (int32) or interval (string), not both.
optional int32 interval_msec = 5;
optional string interval = 16;
```

manugarg (commented):
A few minor comments. Looks mostly good.

probes/options/options.go
nvidyala: Done.

probes/options/options.go
nvidyala: Done.

probes/options/options.go
nvidyala: Done.

probes/proto/config.proto
nvidyala: Done.

probes/proto/config.proto
```diff
@@ -43,11 +43,17 @@ message ProbeDef {
   // this probe only if machine's hostname matches this value.
   optional string run_on = 3;
 
-  // Interval between two probes
-  optional int32 interval_msec = 4 [default = 2000];
+  // Interval between two probes, int32 representing milliseconds
```
manugarg, line 46: Period at the end. Can you please mention that only one of "interval" and "interval_msec" should be defined. I'd structure it this way:

// Interval between two probe runs in milliseconds.
// Only one of "interval" and "inteval_msec" should be defined.
optional int32 interval_msec

// Interval between two probe runs in string format, e.g. 10s.
// Only one of "interval" and "inteval_msec" should be defined.
optional string interval

probes/options/options.go
```diff
@@ -113,14 +115,39 @@ func getSourceIPFromConfig(p *configpb.ProbeDef, l *logger.Logger) (net.IP, erro
...
+		intervalDuration = time.Duration(p.GetIntervalMsec()) * time.Millisecond
+	} else if p.GetInterval() != "" {
+		intervalDuration, err = time.ParseDuration(p.GetInterval())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse interval (%s): %v", p.GetInterval(), err)
+		}
+	}
+
+	if p.GetTimeoutMsec() != 0 && p.GetTimeout() != "" {
+		return nil, fmt.Errorf("both timeout: (%s) and timeout_msec: (%d) are specified", p.GetTimeout(), p.GetTimeoutMsec())
```
manugarg, line 134: Same as above. I'd skip the semicolon after timeout and timeout_msec.

manugarg (commented):
Last couple of comments I think.

probes/proto/config.proto
nvidyala: Done.

probes/options/options.go
nvidyala: Done.

probes/options/options.go
```diff
@@ -113,14 +115,39 @@ func getSourceIPFromConfig(p *configpb.ProbeDef, l *logger.Logger) (net.IP, erro
...
+	}
+
+	if p.GetTimeoutMsec() != 0 && p.GetTimeout() != "" {
+		return nil, fmt.Errorf("both timeout (%s) and timeout_msec (%d) are specified", p.GetTimeout(), p.GetTimeoutMsec())
+	} else if p.GetTimeoutMsec() != 0 {
+		timeoutDuration = time.Duration(p.GetTimeoutMsec()) * time.Millisecond
+	} else if p.GetTimeout() != "" {
+		timeoutDuration, err = time.ParseDuration(p.GetTimeout())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse interval (%s): %v", p.GetTimeout(), err)
```
manugarg, line 140: s/interval/timeout: failed to parse timeout

probes/proto/config.proto
```diff
@@ -43,11 +43,21 @@ message ProbeDef {
   // this probe only if machine's hostname matches this value.
   optional string run_on = 3;
 
-  // Interval between two probes
-  optional int32 interval_msec = 4 [default = 2000];
+  // Interval between two probe runs in milliseconds.
+  // Only one of "interval" and "inteval_msec" should be defined.
+  optional int32 interval_msec = 4;
```
manugarg, line 48: Sorry, I missed this earlier. Now that we don't have an explicit default here, can you add a comment saying: Default interval is 2s (both -- under interval and interval_msec). Same for timeout fields.

probes/options/options.go
nvidyala: Oops, fixed.

probes/proto/config.proto
nvidyala: Updated.

probes/proto/config.proto
```diff
@@ -43,11 +43,25 @@ message ProbeDef {
...
+
+  // Interval between two probe runs in string format, e.g. 10s.
+  // Only one of "interval" and "inteval_msec" should be defined.
+  // Default interval is 2s
+  optional string interval = 16;
+
+  // Timeout for each probe in milliseconds
+  // Only one of "timeout" and "timeout_msec" should be defined.
+  // Default interval is 1s
```
manugarg, line 58:
```suggestion
  // Default interval is 1s
```
timeout. Fix below as well.

nvidyala: Sorry about that, fixed.

probes/proto/config.proto
```diff
@@ -43,11 +43,25 @@ message ProbeDef {
...
   optional string run_on = 3;
 
-  // Interval between two probes
-  optional int32 interval_msec = 4 [default = 2000];
-
-  // Timeout for each probe
-  optional int32 timeout_msec = 5 [default = 1000];
+  // Interval between two probe runs in milliseconds.
+  // Only one of "interval" and "inteval_msec" should be defined.
+  // Default interval is 2s
```
manugarg, line 48:
```suggestion
  // Default interval is 2s.
```

probes/proto/config.proto
```diff
@@ -43,11 +43,25 @@ message ProbeDef {
...
+  optional int32 interval_msec = 4;
+
+  // Interval between two probe runs in string format, e.g. 10s.
+  // Only one of "interval" and "inteval_msec" should be defined.
+  // Default interval is 2s
```
manugarg, line 53:
```suggestion
  // Default interval is 2s.
```

probes/proto/config.proto
```diff
@@ -43,11 +43,25 @@ message ProbeDef {
...
+  optional string interval = 16;
+
+  // Timeout for each probe in milliseconds
+  // Only one of "timeout" and "timeout_msec" should be defined.
+  // Default timeout is 1s
```
manugarg, line 58:
```suggestion
  // Default timeout is 1s.
```

probes/proto/config.proto
```diff
@@ -43,11 +43,25 @@ message ProbeDef {
...
+  optional int32 timeout_msec = 5;
+
+  // Timeout for each probe in string format, e.g. 10s.
+  // Only one of "timeout" and "timeout_msec" should be defined.
+  // Default timeout is 1s
```
manugarg, line 63:
```suggestion
  // Default timeout is 1s.
```

manugarg (approved):

PR #31, opened 2021-11-17
Allow DisableJitter=true and the use of the gRPC service to add probes.

cstahl1:
Fixes bug where if DisableJitter=true, then Prober.Start will exit prematurely.


manugarg (approved):
Thank you so much for fixing it! :)

PR #33, opened 2021-11-30
Allow multiple instances of cloudprober to run sequentially.

cstahl1:
This is most useful for tests in which a full cloudprober is
required. Each test initializes a new instance, calling
InitFromConfig and Start and when the context passed to Start is done,
the cloud prober instance will shut down fully. There is currently no
good way to ensure that the last instance has shut down before starting
a new one, so I used a time.Sleep.

- Added ClearDefaultGRPCServer() to runconfig to clear the saved gRPC
server
- On context close, Start will fully clean up the global variables found
in cloudprober.go
- Added test that starts CloudProber, stops it, and starts it again.
Before this code change, that would have always failed with the second
config ignored.

### DIFF
--- a/cloudprober_test.go
+++ b/cloudprober_test.go
@@ -200,11 +200,12 @@ func TestRestart(t *testing.T) {
 				t.Fatalf("InitFromConfig(ctx, %v): %v", string(b), err)
 			}
 			Start(ctx)
-			ctxTO, cancelTO := context.WithTimeout(ctx, time.Second*30)
-			defer cancelTO()
 
+			// Wait for results from the surfacer, or 30s,
+			// whichever comes first. Since the export rate is 1s,
+			// we should expect results well before 30s has passed.
 			select {
-			case <-ctxTO.Done():
+			case <-time.After(time.Second * 30):
 				t.Fatal("surfacer timed out before getting results")
 			case <-s.c:
 			}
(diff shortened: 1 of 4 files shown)


manugarg:
Thanks @cstahl1! This mostly looks good to me. I'll take a more detailed look later today.

cloudprober_test.go
```diff
@@ -84,3 +96,118 @@ func TestGetDefaultServerPort(t *testing.T) {
...
+			err = InitFromConfig(string(b))
+			if err != nil {
+				t.Fatalf("InitFromConfig(ctx, %v): %v", string(b), err)
+			}
+			Start(ctx)
+			ctxTO, cancelTO := context.WithTimeout(ctx, time.Second*30)
+			defer cancelTO()
+
+			select {
+			case <-ctxTO.Done():
```
manugarg, line 207: It took me a while to understand it. I think this section will benefit from a comment.

```
Start(ctx)

// Before returning, wait for data on the surfacer channel, for up to 30s. 
// Since cloudprober is configured to export data every 1s (see above), we should
// see some data within 30s.
ctxTo, cancelTo := .. 
```

(Please keep the above comment limited to 80 chars).

manugarg (commented):
Thanks for this change, @cstahl1!

One minor comment about adding a comment to clarify :)

cloudprober_test.go
cstahl1: I added a comment, but also changed it from a context.WithTimeout to a time.After to make it more clear what was happening.

PR #35, opened 2021-12-07
Generate a new DNS transaction ID for each probe

cbroglie:
This is more representative of real traffic, which should not repeat
transaction IDs.

### DIFF
--- a/probes/dns/dns.go
+++ b/probes/dns/dns.go
@@ -75,9 +75,10 @@ type Probe struct {
 	l    *logger.Logger
 
 	// book-keeping params
-	targets []endpoint.Endpoint
-	msg     *dns.Msg
-	client  Client
+	targets   []endpoint.Endpoint
+	queryType uint16
+	fqdn      string
+	client    Client
 }
 
 // probeRunResult captures the results of a single probe run. The way we work with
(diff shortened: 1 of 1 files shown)


probes/dns/dns.go
```diff
@@ -133,17 +132,15 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 	}
 	p.updateTargets()
 
-	// I believe these objects are safe for concurrent use by multiple goroutines
-	// (although the documentation doesn't explicitly say so). It uses locks
-	// internally and the underlying net.Conn declares that multiple goroutines
-	// may invoke methods on a net.Conn simultaneously.
-	p.msg = new(dns.Msg)
 	queryType := p.c.GetQueryType()
```
manugarg, line 135: Since we validate query_type here, I think it will be better to store query type in Init() itself. So basically let's do something like:

```
if queryType == configpb.QueryType_NONE || int32(queryType) >= int32(dns.TypeReserved) {
		return fmt.Errorf("dns_probe(%v): invalid query type %v", name, queryType)
}
p.queryType = unit16(queryType)
```

cbroglie: 👍

manugarg (approved):

cbroglie:
Thanks!

manugarg:
> Thanks!

Thank you! :)

Issue #93, opened 2022-02-26
HTTP probes don't refresh bearer tokens


jtse:
**Describe the bug**
My probes against a Google API were initially succeeding but failed with a 401 after an hour. Upon further debugging, I realized:

*  The [HTTP request is created once](https://github.com/cloudprober/cloudprober/blob/master/probes/http/http.go#L443) and used _repeatedly_ for each probe request.
*  The bearer token header is set at the time of HTTP request creation.
*  So in effect, the same bearer token is used over and over. If it bearer token is short-lived like an access token, then the probe will fail when the token expires.

Practically speaking, cloudprober can't be used against many web APIs until this bug is fixed.

Note that cloudprober _is_ getting new access tokens but just isn't using them in the HTTP request. I was able to fix this by simply creating a new HTTP request instance for each probe request.

**Cloudprober Version**
0.11.4 but I suspect it affects the past few versions.

**To Reproduce**
1. Decide on a Google API to test.
2. Open Google Cloud's Cloud Shell (shouldn't cost any).
3. git clone and compile cloudprober.
4. Create a cloudprober.cfg similar to the following:
```
probe {
  name: "foo"
  type: HTTP
  targets {
    host_names: "foo.googleapis.com"
  }
  http_probe {
    relative_url: "/path/to/api/endpoint
    protocol: HTTPS
    method: POST
    headers {
      name: "content-type"
      value: "application/json"
    }
    // Will auth with the Cloud Shell user's access creds.
    oauth_config {
      google_credentials {
      }
    }
  }
  interval_msec: 30000
  timeout_msec: 1000
  validator {
    name: "status_code_200"
    http_validator {
      success_status_codes: "200"
    }
  }
}
```
5. Run cloudprober in Cloud Shell.

Expected: 200
Actual: 200 and then 401s when the token expires.

manugarg:
Thanks for the bug report @jtse. I'll take a look at it.

manugarg:
I see, so we periodically get the new token here:
https://github.com/cloudprober/cloudprober/blob/1dfdd012c8efbad0695b717498ad8f46f9021055/probes/http/http.go#L595

But we recreate HTTP request only if we're resolving first (i.e. `resolve_first` is enabled):
https://github.com/cloudprober/cloudprober/blob/1dfdd012c8efbad0695b717498ad8f46f9021055/probes/http/http.go#L466

[...]

PR #94, opened 2022-02-26
Fix the bug in OAuth handling.

manugarg:
* Fix a bug where we were not really refreshing bearer token in HTTP request (#93).

* Stop storing OAuth bearer token in probe's state, and stop updating it in the target refresh loop. This may create concurrency issues.

* Instead of the above,  get a new token while creating the HTTP request. Note that most OAuth token sources already implement caching so it should be okay.


manugarg:
#95 is a better a fix I think. It's more deterministic in the sense that token is accessed (and refreshed if needed) right before making the HTTP request.

jtse (approved):

manugarg:
Closing it in favor of #95.

Issue #88, opened 2022-02-17
Probes status in Cloudprober UI 


manugarg:
Currently Cloudprober runs probes, and surfaces the generated data (success, failure, latency, etc) to other metrics systems like Prometheus, Cloudwatch, etc. It itself doesn't expose data directly to the users in a way that is easy to interpret. It will be nice if it did that.

manugarg:
Cloudprober now already provides a way to configure surfacers to export probes data as GAUGE metrics. We can possibly implement a surfacer that collects the metrics internally (in-memory), and makes them available through a readable UI. For example, something like this:

|Probe | Status |
|------|-------|
|probe1 | GGRRGRGGG|
|probe2| RRGGGGGGG|

G,R stand for green and red color boxes.

Few things to think about:
* How much data to store, i.e. how long back do we go. This will directly impact the memory usage.
* Do we show latency at all?
* Should color boxes above be clickable, and take one to logs for that probe, does that mean we need to store logs as well.
* etc.

manugarg:
> How much data to store, i.e. how long back do we go. This will directly impact the memory usage.

I think we can keep the probe's summary since the start of Cloudprober. Summary could be like:
Failure ratio in the last: 10m, 1h, 6h, 12h, 1d, 7d, 30d, 1y

manugarg:
I forgot that we'll need to include targets in above presentation. So it will likely look like:

|Probe | Target | Status| Summary (1h) | 12h | 1d | 7d | 30d | 1y |
|------|--------|-------|---------------|-----|----|----|----|---|
|probe1 | target1 | GGRRGRGGG| 99% | 99.7% | 99.6%| 99.5% | 99.6% |99.4%|
|probe1 | target2 |RRGGGGGGG| 98% | 97.9% | 97.5% | 97.2% | 96.9% | 98% |
|probe2 | target | GGGGGGGGG| 100% | 99.97% | 97.5% | 97.2% | 96.9% | 98% |

manugarg:
Just realized, we can show this summary on terminal as well.

An example program using [termbox-go](https://github.com/nsf/termbox-go) to show live ping results on terminal:
Demo: https://github.com/hirose31/pinger
Code: https://github.com/hirose31/pinger/blob/master/pinger.go

manugarg:
This is how it looks like at present:

[...]

PR #105, opened 2022-03-16
[WIP] First draft of probestatus surfacer.

manugarg:
See #88 for some background.


manugarg:
There is still a lot of work a left to be done:

* Truncate list of targets, if there are more than 10-20 targets.
* Show probe success ratio graph using some javascript library.

But I think it's useful already, and it's completely optional right now.

PR #107, opened 2022-03-16
Add support for expanding GCE global forwarding rule targets.

### LINKED ISSUE #106: Support for discovering GCP global forwarding rules targets
Currently cloudprober can only discover forwarding rules that are tied to a region (https://github.com/cloudprober/cloudprober/blob/master/targets/gce/proto/config.proto#L53-L63). I need to probe targets that are configured as global forwarding rules. For that cloudprober needs the ability to discover global forwarding rules.

muntakimsadik:
This adds a new boolean proto field `global_rule` for specifying forwarding rules. This should allow to write probes that  probes global forwarding rules.

resolves #106


manugarg:
Sorry for taking so long to review this.

Ideally we should consolidate the forwarding rules logic here:
https://github.com/cloudprober/cloudprober/blob/master/rds/gcp/forwarding_rules.go

However, as that will require more careful thinking, I am okay with going ahead with this change for now.

manugarg (approved):

PR #108, opened 2022-03-16
[probestatus] Show version, startup time as well.

manugarg:
= In preparation to merge /status and /probestatus, show  version, startup time on /probestatus page.
= Other minor style tweaks.


manugarg:
Another change for #88.

PR #125, opened 2022-03-30
[ping] Add a real ping test.

manugarg:
= This test will run only if ENABLE_EXTERNAL_TESTS environment variable is set.

= Since Github doesn't support ping and IPv6, we'll have to run these tests locally for now.


manugarg:
I think this PR addresses this too: https://github.com/google/cloudprober/issues/227.

PR #128, opened 2022-03-31
Update go.yml

manugarg:


manugarg:
I think this PR addresses this too: https://github.com/google/cloudprober/issues/227.

PR #132, opened 2022-04-06
[probestatus] Add graphs to the probestatus page \o/.

manugarg:
= Very excited about this change. Probestatus page will now show success
ratio graphs.

= There will be one graph per probe, and graph lines will correspond to
the targets.

= We use timeseries compression (values and frequencies) for the graph data to
minimize memory allocation in Cloudprober at the time of graph building. This also
helps the probestatus page size stay low (more values !=> bigger size).

= Graphs also support the following URL params: `graph_endpoint` and
`graph_duration` to move around in the graph. You can set these params only in the
 URL right now, I'll add HTML inputs in a later change.

= We are currently using C3 JS for graphing, and shipping all the
required JS in Cloudprober itself. We'll re-evaluate both the decisions
in future.


manugarg:
This is how graphs will look like on the 'probestatus' page: 

image

PR #153, opened 2022-06-04
aws sdk v1 -> v2

robpickerill:
replace references to the aws sdk v1 with the v2 release and accomodates for the api changes between the two versions


robpickerill:
Was looking into that test failure, it is reproducible on master locally here (might not be :

```
E0605 13:25:48.710239    4729 sysvars.go:64] [nil] Sysvars map is un-initialized. sysvars.Vars() was called before sysvars.Init().
    grpc_test.go:242: Probe results: 1654431954 labels=ptype=grpc,probe=grpc-connectfail,dst=localhost:9 total=3 success=0 =0.000 connecterrors=3
    grpc_test.go:251: Got total=3 connectErrs=3, expecting at least 4 for each
    grpc_test.go:242: Probe results: 1654431960 labels=ptype=grpc,probe=grpc-connectfail,dst=localhost:9 total=9 success=0 =0.000 connecterrors=9
--- FAIL: TestConnectFailures (18.00s)
```

I wasn't able to position the root change that caused it, will circle back round it.

surfacers/cloudwatch/proto/config.proto
```diff
@@ -11,14 +11,14 @@ message SurfacerConf {
   // The cloudwatch resolution value, lowering this below 60 will incur
   // additional charges as the metrics will be charged at a high resolution
   // rate.
-  optional int64 resolution = 2 [default = 60];
+  optional int32 resolution = 2 [default = 60];
```
robpickerill, line 14: not sure if this is best marked as deprecated and I define a new field here, but I can't imagine anyone has a resolution > 2147483647 seconds?

sysvars/sysvars_ec2.go
```diff
@@ -15,43 +15,43 @@
...
 var ec2Vars = func(sysVars map[string]string, tryHard bool, l *logger.Logger) (bool, error) {
-	cfg := &aws.Config{}
-	// If not trying hard (cloud_metadata != ec2), use shorter timeout.
-	if !tryHard {
-		cfg.MaxRetries = aws.Int(0)
-		cfg.EC2MetadataDisableTimeoutOverride = aws.Bool(true)
-		cfg.HTTPClient = &http.Client{
-			Timeout: 100 * time.Millisecond,
+	ctx := context.Background()
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRetryer(func() aws.Retryer {
```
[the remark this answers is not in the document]
robpickerill: The type signature for config.LoadDefaultConfig is variadic over a list of functions that take `*config.LoadOptions`. 

Here `cfg, err := config.LoadDefaultConfig(ctx, opts)` we would be passing a slice of `config.LoadOptionsFunc` and I can't see a way to get to the invocation of `*config.LoadOptions` without invoking the functions outside of the LoadDefaultConfig function.

What about something like this instead, mutating the config struct that is return from LoadDefaultConfig?

```
	ctx := context.Background()

	// isTryHard is a switch over whether we should wait longer, and retry attempts, when communicating
	// with AWS APIs. This is to reduce cloudprober start up time when not running in an AWS
	// environment
	if !tryHard {
		var cancel context.CancelFunc
		ctx, cancel = context.WithTimeout(ctx, 100*time.Millisecond)
		defer cancel()
	}

	cfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		l.Warningf("sysvars_ec2: failed to load default config: %v", err)
		return false, nil
	}

	// patch RetryMaxAttempts with 0 to disable retries when tryHard is false
	if !tryHard {
		cfg.RetryMaxAttempts = 0
	}
```

sysvars/sysvars_ec2.go
```diff
@@ -15,43 +15,51 @@
...
-	md := ec2metadata.New(s)
-	// Doing the availability check in module since we need a session
-	if md.Available() == false {
-		return false, nil
-	}
+	client := imds.NewFromConfig(cfg)
 
-	id, err := md.GetInstanceIdentityDocument()
+	id, err := client.GetInstanceIdentityDocument(ctx, &imds.GetInstanceIdentityDocumentInput{})
+	// the premise behind the error handling here, is that we want to evaluate if we are running in ec2 or not.
```
manugarg, line 59:
```suggestion
	// The premise behind the error handling here, is that we want to evaluate
	// if we are running in ec2 or not.
```

sysvars/sysvars_ec2.go
```diff
@@ -15,43 +15,51 @@
...
+// loadAWSConfig will evaluate if we should apply retries or not to the AWS config
+func loadAWSConfig(ctx context.Context, tryHard bool) (aws.Config, error) {
+	retryMaxAttempts := 3
+
+	// if we are not trying hard, then we can disable any retries
+	if !tryHard {
+		retryMaxAttempts = 0
+	}
+
+	return config.LoadDefaultConfig(ctx, config.WithRetryMaxAttempts(retryMaxAttempts))
```
manugarg, line 37: Can we specifying max retries for tryHard case.. how about just do this:

```
if !tryHard {
   return config.LoadDefaultConfig(ctx, config.WithRetryMaxAttempts(0))
}
return config.LoadDefaultConfig(ctx)
```

sysvars/sysvars_ec2.go
```diff
@@ -15,43 +15,51 @@
...
+}
+
 var ec2Vars = func(sysVars map[string]string, tryHard bool, l *logger.Logger) (bool, error) {
-	cfg := &aws.Config{}
-	// If not trying hard (cloud_metadata != ec2), use shorter timeout.
+	ctx := context.Background()
+
+	// if we are not trying hard, then we can apply a strict timeout on the AWS API calls
```
manugarg, line 43: Previous comment probably added more color to try hard :)
```suggestion
	// If not trying hard (cloud_metadata != ec2), use shorter timeout.
```

manugarg (approved):
Sorry for the super long delay, Rob.

This change looks good to me. Last couple of minor suggestions.

robpickerill:
no apologies needed, and no problems at all, let me finalise the last few changes - aiming for next few days but latest will be the weekend.

sysvars/sysvars_ec2.go
robpickerill: your right, honestly, I overlooked the previous comments and forgot them - my oversight

robpickerill:
Tests fail, as `cfg.RetryMaxAttempts` is set to 0 as default. Looking at the implementation of the client, it looks like the retry attempts value is stored under the retryer here: https://github.com/aws/aws-sdk-go-v2/blob/main/feature/ec2/imds/api_client.go#L75-L77 and if the retryer is set in the client then MaxRetryAttempts is ignored: https://github.com/aws/aws-sdk-go-v2/blob/main/config/load_options.go#L52-L56

sysvars/sysvars_test.go
```diff
@@ -120,3 +122,77 @@ func TestInitCloudMetadata(t *testing.T) {
...
+			}
+
+			if tc.retryOptions.retryerAvailable {
+				if cfg.Retryer().MaxAttempts() != tc.retryOptions.retryCount {
+					t.Errorf("retry test: %s, retry count expected: %d, got: %d", name, tc.retryOptions.retryCount, cfg.Retryer().MaxAttempts())
+				}
+			}
+		})
+	}
+}
```
manugarg, line 198: Can we please flatten out the test structure like this (you can also just accept this commit suggestion but formatting will likely off):

```suggestion
	tests := map[string]struct{
	        tryHard               bool
		retryerAvailable bool
           	retryCount          int
	}{
		"notTryingHard": {
			tryHard:          false,
		    	retryerAvailable: true,
		    	retryCount:       1,
		},
		"tryingHard": {
		    	tryHard:	 false,
		    	retryerAvailable: false,
		    	retryCount:       0,
		},
	}
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			ctx := context.Background()
			cfg, err := loadAWSConfig(ctx, tc.tryHard)
			if err != nil {
				t.Error(err)
			}

 			if tc.retryerAvailable {
				if cfg.Retryer().MaxAttempts() != tc.retryCount {
					t.Errorf("retry test: %s, retry count expected: %d, got: %d", name, tc.retryCount, cfg.Retryer().MaxAttempts())
				}
			}
		})
	}
}
```

sysvars/sysvars_test.go
```diff
@@ -120,3 +122,77 @@ func TestInitCloudMetadata(t *testing.T) {
...
+			if tc.timeoutOptions.timeoutAvailable {
+				select {
+				case <-ctx.Done():
+				case <-time.After(tc.timeoutOptions.timeoutDuration):
+					t.Errorf("timeout did not occur in: %s", tc.timeoutOptions.timeoutDuration)
+				}
+			}
+		})
+	}
+}
```
manugarg, line 158: Honestly I don't think we need a test for the context itself. If anything, a test verifying that EC2 vars initialization (end-to-end) never takes more than 100ms will be more useful, whether it succeeds or not, but let's do that separately. (Or, I could implement a test to verify that cloudprober always starts within 1s or something like that).

sysvars/sysvars_ec2.go
```diff
@@ -66,3 +54,37 @@ var ec2Vars = func(sysVars map[string]string, tryHard bool, l *logger.Logger) (b
...
+
+// createContext will create a new context, and if tryHard is false then
+// use a shorter timeout
+func createContext(tryHard bool) (context.Context, context.CancelFunc) {
+	if !tryHard {
+		return context.WithTimeout(context.Background(), time.Millisecond*100)
+	}
+
+	return context.WithCancel(context.Background())
+}
```
manugarg, line 66: How will you feel about getting rid of this function and just doing this in the main function itself. As I commented below, I think we can skip testing context timeout.

sysvars/sysvars_ec2.go
```diff
@@ -66,3 +54,37 @@ var ec2Vars = func(sysVars map[string]string, tryHard bool, l *logger.Logger) (b
...
+// createAWSIMDSClient creates a new IMDS client, with retries enabled if tryHard is true,
+// and retries disabled if tryHard is false
+func createAWSIMDSClient(ctx context.Context, tryHard bool) (*imds.Client, error) {
+	cfg, err := loadAWSConfig(ctx, tryHard)
+	if err != nil {
+		return nil, err
+	}
+
+	return imds.NewFromConfig(cfg), nil
+}
```
manugarg, line 77: Since this function is just passing along tryHard to loadAWSConfig.. I think we can get rid of this function as well. In ec2Vars just get the config with tryHard using loadAWSConfig and later on just use config to get the IMDS client.

manugarg (commented):
Sorry, for the back and forth. Appreciate your patience.

robpickerill:
> Sorry, for the back and forth. Appreciate your patience.

No problems, I appreciate the time, and I understand I've refactored on the PR a few times. Context on the the two functions is that I found that the retries were still active under the previous implementation, and so I could address the items tryHard did in a small scope I pushed them into functions, so I could mentally tick them off so I didn't drop any functionality in the move to v2.

No issues with any of the comments, let me address comments and get the PR patched up for you.

PR #162, opened 2022-06-16
Fix cluster role on how to run k8s section

spirosoik:


manugarg:
Thank you, @spirosoik!

PR #185, opened 2022-09-16
Feature/postgres labels column

jessequinn:
This feature allows an user to select which labels will be columns in Postgres rather than simply creating a jsonb representation of labels.


jessequinn:
```create table metrics
(
    time         timestamp with time zone not null,
    metric_name  text                     not null,
    value        double precision,
    labels       jsonb,
    label__dst   text,
    label__code  text,
    label__probe text,
    label__ptype text
);
```

The jsonb can be dropped if using the `label_to_columns`

```
surfacer {
  type: POSTGRES

  postgres_surfacer {
    connection_string: "postgresql://postgres:xxxxx@localhost:5432/cloudprober?sslmode=disable"
    metrics_table_name: "metrics"
    metrics_buffer_size: 120000

    label_to_column {
      label: "probe"
      column: "label__probe"
    }

    label_to_column {
      label: "dst"
      column: "label__dst"
    }

    label_to_column {
      label: "code"
      column: "label__code"
    }

    label_to_column {
      label: "ptype"
      column: "label__ptype"
    }
  }
}
```
An example select:

Screen Shot 2022-09-16 at 09 26 37

surfacers/postgres/postgres.go
```diff
@@ -184,26 +182,47 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
-	}
-
-	for _, pgMetric := range emToPGMetrics(em) {
-		var s string
-		if s, err = labelsJSON(pgMetric.labels); err != nil {
+	// Transaction for defined columns
+	if len(s.c.GetLabelToColumn()) > 0 {
+		// Prepare a statement to COPY table from the STDIN.
+		stmt, err = txn.Prepare(pq.CopyIn(s.c.GetMetricsTableName(), generateColumns(s.c.GetLabelToColumn())...))
+		if err != nil {
```
manugarg, line 197: generateColumns can be called at the time of initialization, i.e. inside the init(). Just call it there and keep the generated columns list as a surfacer field:

in init()..

```
if len(s.c.GetLabel... ) > 0 {
       s.columns = append([]string{"time..}, make([]string, len(s.c.Get..)...)
       for i, v := range ltc {
	     s.columns[i+3] = v.GetColumn())
	}
}
```

surfacers/postgres/postgres.go
```diff
@@ -184,26 +182,47 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
 			return err
 		}
-		if _, err = stmt.Exec(pgMetric.time, pgMetric.metricName, pgMetric.value, s); err != nil {
+
+		for _, pgMetric := range emToPGMetrics(em) {
+			mtrs := []interface{}{pgMetric.time, pgMetric.metricName, pgMetric.value}
```
manugarg, line 202: It's not clear to me what mtrs stand for? Can you add a comment or maybe a different name?

surfacers/postgres/postgres_test.go
```diff
@@ -97,3 +98,76 @@ func isRowExpected(row pgMetric, t time.Time, metricName string, value string, l
 
 	return true
 }
+
+func Test_generateColumns(t *testing.T) {
```
manugarg, line 102: Functions in this file are named in a non-standard way (with respect to Cloudprober at least). Can you please rename this to TestGenerateColumns? Same for below. Thanks!

manugarg (commented):
Added inline comments. Please take a look.

jessequinn:
I added the changes; however, i also changed the surface like so:

```
surfacer {
  type: POSTGRES

  postgres_surfacer {
    connection_string: "postgresql://postgres:xxx@localhost:5432/cloudprober?sslmode=disable"
    metrics_table_name: "metrics"
    metrics_buffer_size: 120000

    labels_to_column {
      label_to_column {
        label: "dst"
        column: "label__dst"
      }
      label_to_column {
        label: "code"
        column: "label__code"
      }
      label_to_column {
        label: "probe"
        column: "label__probe"
      }
      label_to_column {
        label: "ptype"
        column: "label__ptype"
      }
    }
  }
}
```

I felt that it made more send to keep the label_to_column within something else.

jessequinn:
one other thing, DropGoMetrics i added as an example. I was wondering if we could add a flag for that to drop sysvars to the postgres surfacer. I don't want to be flooded with metrics that I don't need.

manugarg:
> I added the changes; however, i also changed the surfacer like so:
> 
> ```
> surfacer {
>   type: POSTGRES
> 
>   postgres_surfacer {
>     connection_string: "postgresql://postgres:xxx@localhost:5432/cloudprober?sslmode=disable"
>     metrics_table_name: "metrics"
>     metrics_buffer_size: 120000
> 
>     labels_to_column {
>       label_to_column {
>         label: "dst"
>         column: "label__dst"
>       }
>       label_to_column {
>         label: "code"
>         column: "label__code"
>       }
>       label_to_column {
>         label: "probe"
>         column: "label__probe"
>       }
>       label_to_column {
>         label: "ptype"
>         column: "label__ptype"
>       }
>     }
>   }
> }
> ```
> 
> I felt that it made more sense to keep the label_to_column within something else.

So In a way, label_to_coloumn is itself a container, and contained within overall conf. Probably the following textproto syntax for repeated fields will make you reconsider this :)

```
label_to_column: [{
  label: "probe"
   column: "label__probe"
}, {
  label: "target"
  column: "label_target"
}]
```

surfacers/postgres/postgres.go
```diff
@@ -228,6 +256,9 @@ func (s *Surfacer) init(ctx context.Context) error {
 	}
 	s.writeChan = make(chan *metrics.EventMetrics, s.c.GetMetricsBufferSize())
 
+	// Generate the desired columns either with 'labels' by default or select 'labels' based on the labels_to_column fields
+	s.generateColumns()
```
manugarg, line 260: Instead of modifying the state, can we simply do this:

```
if s.c.GetLabel.. {
   s.columns = generateColumns(s.c.GetLabelsToCo...())
}
```

jessequinn:
please see new changes as per your requests

surfacers/postgres/postgres.go
```diff
@@ -184,26 +183,48 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
 			return err
 		}
-		if _, err = stmt.Exec(pgMetric.time, pgMetric.metricName, pgMetric.value, s); err != nil {
+
+		for _, pgMetric := range emToPGMetrics(em) {
+			// args are the column values generated based on the chosen labels
+			args := []interface{}{pgMetric.time, pgMetric.metricName, pgMetric.value}
+			for _, v := range generateValues(pgMetric.labels, s.c.GetLabelToColumn()) {
+				args = append(args, v)
+			}
```
manugarg, line 207: I think you could do the following to save on a few appends:

args = append(args, generateValues(pgMetric.labels, s.c.GetLabelToColumn())...)

surfacers/postgres/postgres.go
```diff
@@ -184,26 +183,48 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
+		stmt, err = txn.Prepare(pq.CopyIn(s.c.GetMetricsTableName(), s.columns...))
+		if err != nil {
 			return err
 		}
+
+		for _, pgMetric := range emToPGMetrics(em) {
+			var s string
+			if s, err = labelsJSON(pgMetric.labels); err != nil {
+				return err
+			}
```
manugarg, line 223: To be honest, code for both paths looks quite similar, for example, line 214 to 220 are same for both paths. For loop is also same, you just do different things with labels. What do you think?

manugarg (commented):
Looks mostly good. Last couple of suggestions I missed last time.

surfacers/postgres/postgres_test.go
jessequinn: agreed

surfacers/postgres/postgres.go
jessequinn: fixed

surfacers/postgres/postgres.go
jessequinn: agreed

surfacers/postgres/postgres.go
```diff
@@ -184,25 +183,40 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
+
 	// Begin a transaction.
 	txn, err := s.db.Begin()
 	if err != nil {
 		return err
 	}
 
 	// Prepare a statement to COPY table from the STDIN.
-	stmt, err := txn.Prepare(pq.CopyIn(s.c.GetMetricsTableName(), "time", "metric_name", "value", "labels"))
+	stmt, err = txn.Prepare(pq.CopyIn(s.c.GetMetricsTableName(), s.columns...))
```
manugarg, line 195: Super minor nit: you can skip declaring stmt above now. Just use ":" here.

surfacers/postgres/postgres.go
```diff
@@ -184,25 +183,40 @@ func New(ctx context.Context, config *configpb.SurfacerConf, l *logger.Logger) (
...
 	if err != nil {
 		return err
 	}
 
-	for _, pgMetric := range emToPGMetrics(em) {
-		var s string
-		if s, err = labelsJSON(pgMetric.labels); err != nil {
-			return err
+	// Transaction for defined columns
+	if len(s.c.GetLabelToColumn()) > 0 {
```
manugarg, line 201: Minor nit, feel free to ignore:

We could move the labelToColumn check inside and keep most of the loop body same. Only downside is cost of additional check in the loop, but readability benefits probably outweigh here.

manugarg (approved):
Couple of minor nits/suggestions. Looks good otherwise. I am accepting this. Please feel free to address the comments as you see fit.

surfacers/postgres/postgres.go
jessequinn: resolved

jessequinn:
In addition, I have another PR to do for postgres. I want to include connection limits. we have noticed a lot of connections being made based on postgres surfacer. easily 100 connections can be made.. but this then blocks other connections.

manugarg:
> once approved and merged, how long before you will release it as a tag?

For non-emergency changes, I usually wait for 1-2 weeks before a formal release. You can use the "master" tag in the meantime.

> In addition, I have another PR to do for postgres. I want to include connection limits. we have noticed a lot of connections being made based on postgres surfacer. easily 100 connections can be made.. but this then blocks other connections.

One suggestion, in the very beginning of this surfacer, was to batch Postgres calls, i.e. push multiple metrics in one commit. That should help with multiple connections. It can be done with a cache and a timer, i.e. cache the metrics as they come in and push them on a timer. Stackdriver and Clouswatch surfacers do that:
https://github.com/cloudprober/cloudprober/blob/master/surfacers/stackdriver/stackdriver.go

Issue #197, opened 2022-10-21
[probes.http] Automatically decide if we should resolve_first


manugarg:
For example, there is no point in not using resolve_first for k8s endpoints targets. See #196 for more details.

I guess we can add a label to resources to signify their type and use that label to resolve first, or we could add a generic features label to such resources. For example:

```
labels[_cloudprober_features] = 'resolve_first,..something_else...'
```

manugarg:
#198 and #199 should take care of it.

manugarg:
I've tested it myself, over a local k8s setup, that this works.

Issue #196, opened 2022-10-21
[k8s rds] Resources discovered on k8s use made up names and not IPs for probing


pugmajere:
**Describe the bug**

Discovering endpoint resources on k8s for HTTP probes results in errors like this:

W1021 20:59:13.816516       1 http.go:307] [cloudprober.sdc-service-edge] Target:edge_10.10.88.28_service-discovery, URL:https://edge_10.10.88.28_service-discovery:9090/ready, http.doHTTPRequest: Get "https://edge_10.10.88.28_service-discovery:9090/ready": dial tcp: lookup edge_10.10.88.28_service-discovery on 169.254.20.10:53: no such host

This name isn't resolvable, and appears to be made up by Cloudprober. There don't appear to be any options to just use the discovered IP, rather than the resource name. ("resource" in this case refers to the internal Cloudprober concept of "endpoint resources", and not anything directly k8s related, as far as I can tell.)

**Cloudprober Version**

v0.11.9 I think.

**To Reproduce**
(slightly simplified config; I dropped some filters and some of the templates being expanded here.)

```
probe {
    type: HTTP
    name: "my_service"
    targets {
        rds_targets {
            resource_path: "k8s://endpoints"
        }
    }
    interval: "30s"
    timeout: "5s"
    {{template "standard_latency_options" .}}

    http_probe {
        protocol: HTTPS
        relative_url: "/ready"
        port: 9090
    }

    validator {
        name: "status_code_2xx"
        http_validator {
            success_status_codes: "200-299"
        }
    }

    validator {
        name: "OK?"
        regex: "LIVE"
    }

}

rds_server {
  provider {
    kubernetes_config {
      endpoints {}
    }
  }
}

```

manugarg:
Actually there is an easy fix for it (we should not even required that fix though, I'll see how we can do that), you just need to add the following field to http_probe stanza:

https://github.com/cloudprober/cloudprober/blob/5128a133a11a9cd51c1d63f02b9131be370a28ae/probes/http/proto/config.proto#L43

pugmajere:
Nice. I'll try that now.

manugarg:
> (we should not even required that fix though, I'll see how we can do that), 

Filed #197 for this.

manugarg:
@pugmajere, Ryan, did it work?

manugarg:
I tested #198 in a local cluster and it's working as expected.

[...]

PR #198, opened 2022-10-22
Store target IP in the endpoint struct itself.

manugarg:
* For discovered resources, store discovered IP address in the endpoint struct itself, and provide a `Resolve` method on Endpoint to use this information if available.
* For HTTP probes, use the above data to decide if we should resolve first. Note that for most other probes we anyway use the discovered IP address. Also, this behavior can be easily disabled by setting resolve_first to false.
* [TODO]: Change other probes to use this method as well.


manugarg:
This is to address #196 and #197.

PR #202, opened 2022-10-26
[probestatus] Move probe status page to /status.

manugarg:
* Earlier default URL, /probestatus, will redirect to the new default URL.
* Old status page has been renamed to /config-parsed and is linked from the new status page.


manugarg:
Closing. Need to re-do this.

PR #205, opened 2022-10-26
add support for setting environment variables for external probes

alex-laties:
Adds the ability to configure environment variables for external probers.

This is useful for golang binaries where configuration is done via environment variables during `init()` functions (anti-pattern, I know, but we can't fix every upstream project to be configurable after init).


probes/external/proto/config.proto
```diff
@@ -26,6 +26,13 @@ message ProbeConf {
   // will get converted to: /tools/recreate_vm -vm ig-us-central1-a
   required string command = 2;
 
+  // EnvVars for the command.
+  message EnvVar {
+    optional string name = 1;
+    optional string value = 2;
+  }
+  repeated EnvVar envVars = 6;
```
manugarg, line 34: Can you name this env_vars? to be consistent with naming elsewhere.

probes/external/external_test.go
```diff
@@ -168,7 +168,19 @@ func runAndVerifyProbe(t *testing.T, p *Probe, tgts []string, total, success map
 	}
 }
 
-func createTestProbe(cmd string) *Probe {
+func createTestProbe(cmd string, envVars map[string]string) *Probe {
+	return createTestProbeWithEnv(cmd, envVars)
+}
```
manugarg, line 173:
```suggestion
func createTestProbe(cmd string) *Probe {
	return createTestProbeWithEnv(cmd, nil)
}
```

probes/external/external_test.go
```diff
@@ -336,16 +349,23 @@ func TestProbeServerLocalPipeClose(t *testing.T) {
 
 func TestProbeOnceMode(t *testing.T) {
 	testCmd := "/test/cmd --arg1 --arg2"
+	testEnv := []string{"foo=bar", "hello=dolly"}
+	testEnvMap := make(map[string]string{})
```
manugarg, line 353:
```suggestion
	testEnvMap := make(map[string]string)
```

probes/external/proto/config.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.5
+// 	protoc        v3.12.4
```
manugarg, line 4: Curious how did you generate this. We (tools/gen_pb_go.sh) should be using 21.5:
https://github.com/cloudprober/cloudprober/blob/5f97a028a6ad243c1b887411cddd53b8d104c9ee/tools/gen_pb_go.sh#L19

Maybe you already had protoc in the path?

manugarg:
@alex-laties I was wondering if you want to follow up on this?

probes/external/proto/config.pb.go
alex-laties: Correct, I was building on a system that already had a version of protoc installed and the `gen_pb_go.sh` does not check the versions to ensure conformance.

To correct this, there are a couple options I see:
 - always download the specified protoc to a temp location, update build chain to always call this temp protoc
   - i'm unclear how many places I would need to explicitly point to this new location, and I think the approach is brittle in the face of new calls to protoc as those places would also have to be correctly prefixed.
 - download the specified protoc to a temp location, mangle PATH so this new version is preferred
   - this is a very old school option with a lot of pitfalls, like failing cleaning up the PATH if builds fail. tooling exists to help here
 - set up a dockerfile for builds that installs the correct version of protoc
   - create something like `Dockerfile.genpb` with the correct versions of the build toolchain installed
   - run `gen_pb_go.sh` script in a container of the genpb environment with the source dir bind-mounted in
   - certainly not immediately obvious to do this, so would likely need updates to documentation and probably a wrapper in the Makefile to make invocation hygenic
   
Of the ways I can think of, I tend to prefer the build container option, but am open to other ideas or preferences here.

alex-laties: I've added the dockerfile and makefile wrapper to give an idea of what I meant.

probes/external/external_test.go
```diff
@@ -342,7 +342,7 @@ func TestProbeOnceMode(t *testing.T) {
 	tgts := []string{"target1", "target2"}
 
 	// Set runCommand to a function that runs successfully and returns a pyload.
-	p.runCommandFunc = func(ctx context.Context, cmd string, cmdArgs []string) ([]byte, error) {
+	p.runCommandFunc = func(ctx context.Context, cmd string, cmdArgs []string, envVars []string) ([]byte, error) {
 		var resp []string
 		resp = append(resp, fmt.Sprintf("cmd \"%s\"", cmd))
 		resp = append(resp, fmt.Sprintf("num-args %d", len(cmdArgs)))
```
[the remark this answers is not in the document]
alex-laties: rather than printing, I verify that the env vars passed in here are what we expect. I also export a new metric (num-env-vars) that reports how many env vars are seen and check if the value is what we expect.

alex-laties:
@manugarg I think this PR is ready for review now.

manugarg:
Hey @alex-laties, I appreciate the suggestions for fixing the protobuf generation part. However, for now, I've done something simpler -- download protoc regardless of whether it's available or not.

Can you please just run the latest [tools/gen_pb_go.sh](https://github.com/cloudprober/cloudprober/blob/master/tools/gen_pb_go.sh) again, and upload the hopefully cleaned up PR.

Sorry for all the trouble, and thanks again for the contribution.

manugarg:
@alex-laties I was wondering if you still want to continue this? If yes, please see my previous comment. Thanks1

PR #209, opened 2022-10-31
[status page] Further improve our status page.

manugarg:
* Add probe selectors to the status page.
   * Reduce drop if no data interval to 6h (from 24h).


manugarg:
image

Issue #219, opened 2022-11-25
Example config cause panic


dawid-remitly:
**Describe the bug**
Running test config from https://cloudprober.org/getting-started/ cause *panic*

**Cloudprober Version** v0.12.1

**To Reproduce**
```
cat > /tmp/cloudprober.cfg <<EOF
probe {
  name: "google_homepage"
  type: HTTP
  targets {
    host_names: "www.google.com"
  }
  interval_msec: 5000  # 5s
  timeout_msec: 1000   # 1s
}
EOF
$ cloudprober -config_file /tmp/cloudprober.cfg
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40 pc=0x1039a3704]

goroutine 45 [running]:
github.com/cloudprober/cloudprober/probes/http.(*Probe).httpRequestForTarget(0x140001f72b0, {{0x1400031a18c, 0xe}, 0x0, {0x0, 0x0, 0x0}, 0x0, {0x0, 0x0, ...}})
	/Users/runner/work/cloudprober/cloudprober/probes/http/request.go:98 +0x104
github.com/cloudprober/cloudprober/probes/http.(*Probe).startForTarget(0x140001f72b0, {0x1042ba3c0, 0x14000456740}, {{0x1400031a18c, 0xe}, 0x0, {0x0, 0x0, 0x0}, 0x0, ...}, ...)
	/Users/runner/work/cloudprober/cloudprober/probes/http/http.go:483 +0xf0
github.com/cloudprober/cloudprober/probes/http.(*Probe).updateTargetsAndStartProbes.func2({{0x1400031a18c, 0xe}, 0x0, {0x0, 0x0, 0x0}, 0x0, {0x0, 0x0, 0x0}}, ...)
	/Users/runner/work/cloudprober/cloudprober/probes/http/http.go:578 +0x138
created by github.com/cloudprober/cloudprober/probes/http.(*Probe).updateTargetsAndStartProbes
	/Users/runner/work/cloudprober/cloudprober/probes/http/http.go:574 +0x53c
```

**Additional context**
tested on darwin-x86_64 and darwin-arm64, same result

manugarg:
Good catch. Last change made http_probe a requires stanza. You can fix it for now by adding an empty http_probe stanza to your probe config. I'll follow up with a fix soon.

vancouverwill:
[...]

PR #223, opened 2023-01-05
[probes.http] Initialize probe config with zero values if config is nil.

manugarg:
…ot given.


manugarg:
This should fix #219.

Issue #206, opened 2022-10-28
Cloudprober on Google Cloud Run


michaelsatanovsky:
I have been trying to set up Cloudprober to run on Google Cloud's Cloud Run from the image listed on docker. However, I've run into issues with it. 

I get a general error:
`The user-provided container failed to start and listen on the port defined provided by the PORT=9313 environment variable.` 

And more specific errors within the logs:
`cloudprober.go:134] Error reading config from metadata. Err: metadata: GCE metadata "project/attributes/cloudprober_config" not defined`

`logger.go:205] [cloudprober.sysvars] Error getting instance name on GCE: metadata: GCE metadata "instance/name" not defined`

`cloudprober.go:182] Error initializing cloudprober. Err: sysvars_gce: error while getting internal_ip from metadata: metadata: GCE metadata "instance/network-interfaces/0/ip" not defined"`

The "[How-Tos - Running on Kubernetes](https://cloudprober.org/how-to/run-on-kubernetes/)" page on Cloudprober's website is helpful for setting up Cloudprober on GKE, but it is a bit overkill for some users (customers of a Google Cloud service). Would it be possible to add similar support for Cloud Run (or point me to some tools to debug this)?

manugarg:
@michaelsatanovsky thanks for reporting this.

There is a workaround for this. You can add the following flag to the command line to disable getting cloud metadata: `--cloud_metadata=none`.

This error is coming from the fact Cloudprober is discovering it's running on Google cloud, but it's not able to get certain metadata variables. On kubernetes we return early: https://github.com/cloudprober/cloudprober/blob/7d1085c3161c24215864dfee984b424076834bd4/sysvars/sysvars_gce.go#L66.

If we can easily identify that we're running on Cloud Run we can make this more automatic. I'll look into that.

michaelsatanovsky:
Hi @manugarg,

Thanks for following up on this! I was unable to get Cloudprober to run on Cloud Run as it seemed to require access to other cloud metadata. I ended up using a VM instance. Should Cloudprober work with Cloud Run after this fix?

Thanks again!

manugarg:
Hello, @michaelsatanovsky 

This change has not been released yet. If you want to test before the release, you can use the docker image tag - "master". Let me know if that doesn't work as well.

manugarg:
I see. Thanks for retrying it. I'll take another look.

manugarg:
Sorry for the delay. I found the issue and submitted the fix:
[...]

PR #225, opened 2023-01-05
[sysvars.gce] Fix Google Cloud Run handling.

manugarg:


manugarg:
This should fix #206.

PR #230, opened 2023-01-20
Fix keep alive for HTTP probes.

darinpeetz:
One commit to move the call to clientsForTarget so new clients aren't created every probing interval, and another to make the MaxIdleConns a configurable option so that keep alive works with more than 256 targets.

### DIFF
--- a/probes/http/http.go
+++ b/probes/http/http.go
@@ -488,6 +488,7 @@ func (p *Probe) startForTarget(ctx context.Context, target endpoint.Endpoint, da
 	ticker := time.NewTicker(p.opts.Interval)
 	defer ticker.Stop()
 
+	clients := p.clientsForTarget(target)
 	for ts := time.Now(); true; ts = <-ticker.C {
 		// Don't run another probe if context is canceled already.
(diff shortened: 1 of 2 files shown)


manugarg:
Looks like this bug appeared in this PR: https://github.com/cloudprober/cloudprober/commit/6a143254a57f377f542b1e256150cdb2df07f8e8

Thanks for catching this @darinpeetz and sending a fix. I'll review it more thoroughly in a day or two.

manugarg:
@darinpeetz, can you please split these two changes into two PRs:
1. Fix the keep-alive issue
2. Add an option for max idle connections.

Also, build failed for the max idle conn change because you've not generated the Go code for the protobuf change. Can you please run this script (https://github.com/cloudprober/cloudprober/blob/master/tools/gen_pb_go.sh) and upload the new diff.

darinpeetz:
I split this into two new PRs and ran the script you mentioned (I knew I was missing something but wasn't sure how to generate the go code, thanks for pointing that out). Closing this one in favor of the newer PRs.

PR #231, opened 2023-01-23
Move clientsForTarget call out of probing loop.

darinpeetz:
The repeated calls to clientsForTarget are generating new clients at every probing interval, which effectively makes the keep alive option useless (a new connection is created for every new client). Pulling that function call out of the loop and storing a single client list (one client per request) allows keep alive to function as normal again.


manugarg:
Thanks once again for this PR. I've added a test to catch regressions in future:
https://github.com/cloudprober/cloudprober/pull/233

manugarg (approved):

manugarg:
Demonstration that tests in #233 will catch such regressions in future:
image

PR #234, opened 2023-01-25
Added custom labels option for cloud log entries

v-pratap:
Added a custom label option in cloudprober's logger.go file for having custom labels in cloud logger entries. Currently we are logging all error message and details in cloud logger and these logs are needed for better debugging experience. And there is no way to get any particular log, so, a probe_id is generated and we need this probe_id in the labels of log entry, so, that we can search any particular log. 

Screenshot of cloud logger after using label option is attached.
cloud logger label

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -150,15 +150,14 @@ func NewCloudproberLog(component string) (*Logger, error) {
 func New(ctx context.Context, logName string, opts ...Option) (*Logger, error) {
 	l := &Logger{
 		name:                logName,
+		labels:              make(map[string]string),
 		debugLog:            enableDebugLog(*debugLog, *debugLogList, logName),
 		disableCloudLogging: *disableCloudLogging,
 	}
 	for _, opt := range opts {
 		opt(l)
 	}
-	if l.labels == nil {
-		l.labels = make(map[string]string)
-	}
+
 	if !metadata.OnGCE() || l.disableCloudLogging {
 		return l, nil
 	}
@@ -170,7 +169,8 @@ func New(ctx context.Context, logName string, opts ...Option) (*Logger, error) {
 	return l, nil
 }
 
-// WithLabels function can be used to set any parameter in Logger struct.
+// WithLabels option can be used to add a set of labels to all logs, e.g.
+// logger.New(ctx, logName, logger.WithLabels(myLabels))
 func WithLabels(labels map[string]string) Option {
 	return func(l *Logger) {
 		l.labels = labels
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -130,6 +130,9 @@ type Logger struct {
 	// metadata on all logging messages.
 }
 
+// Option can be used for adding additional metadata information in logger.
+type Option func(*Logger)
+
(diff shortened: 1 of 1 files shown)


v-pratap:
Hi @manugarg, Can you please take a look at this PR?

manugarg:
@v-pratap Thanks for the submission. I'll take a look at it later in the week.

v-pratap:
@manugarg yes we are using it directly. Previously we were using funciton **"New(ctx context.Context, logName string)"** for initialising the logger, but currently we need some metadata entry in logs like probe_id  and for that we are using this function **"NewCloudLoggerWithLabels(ctx context.Context, logName string, labels map[string]string)"** directly.

manugarg:
I think we should use options pattern for this instead of defining a new function. It will look something like this:

```
func New(ctx context.Context, logName string, opts ...Option) (*Logger, error) {
         ...
        // after creating the logger l
	for _, opt := range opts {
		opt(l)
	}
}

// Define Option
type Option func(*Logger)


func WithLabels(labels map[string]string) Option {
	return func(l *Logger) {
		l.Labels = labels
	}
}

// To add labels, do this:
logger.New(ctx, logName, WithLabels(labels)

// Following calls (existing) will also be valid:
logger.New(ctx, logName)
```

Does it make sense? Here is more documentation on it:
https://levelup.gitconnected.com/options-pattern-in-golang-9a0384a9d8db

logger/logger.go
```diff
@@ -140,24 +147,36 @@ func NewCloudproberLog(component string) (*Logger, error) {
...
 	l := &Logger{
 		name:                logName,
 		debugLog:            enableDebugLog(*debugLog, *debugLogList, logName),
 		disableCloudLogging: *disableCloudLogging,
 	}
-
+	for _, opt := range opts {
+		opt(l)
+	}
+	if l.labels == nil {
```
manugarg, line 159: Minor suggestion:

Why not just initialize l with an empty map at line 151:
```
   l := &Logger{
     name: logName,
     ...
     labels: make(map[string]string),
   }
```

We can then skip the check here.

logger/logger.go
```diff
@@ -140,24 +147,36 @@ func NewCloudproberLog(component string) (*Logger, error) {
...
+	l.Infof("Running on GCE. Logs for %s will go to Cloud (Stackdriver).", l.name)
 
-	l.Infof("Running on GCE. Logs for %s will go to Cloud (Stackdriver).", logName)
 	if err := l.EnableStackdriverLogging(ctx); err != nil {
 		return nil, err
 	}
 	return l, nil
 }
 
+// WithLabels function can be used to set any parameter in Logger struct.
```
manugarg, line 173:
```suggestion
// WithLabels option can be used to add a set of labels to all logs, e.g.
// logger.New(ctx, logName, logger.WithLabels(myLabels))
```

manugarg (commented):
Thanks for implementing the suggestion. Couple of minor suggestions.

v-pratap:
Thanks. I have changed as per your suggestion.

manugarg (approved):

PR #269, opened 2023-02-07
[probes.grpc] add option to use insecure transport credentials

abaratif:
Hello!

I recently added a probe to my team's cloudprober deployment to check a grpc endpoint. Before adding, I confirmed the configuration with a local instance of the application, and a local instance of cloudprober. The configuration seemed fine, but upon deploying into production, the probe errored out.
After some unsuccessful debugging, I set the following env vars for grpc 

```GRPC_GO_LOG_VERBOSITY_LEVEL=99 GRPC_GO_LOG_SEVERITY_LEVEL=info```
and was presented with the following error:

```Err: connection error: desc = "transport: authentication handshake failed: local credentials rejected connection to non-local address```

I found that this error was from the [credentials/local pkg](https://github.com/grpc/grpc-go/blob/master/credentials/local/local.go#L63-L73) of go-grpc.

Since the endpoint I was trying to hit does not use tls, but isn't local, I patched cloudprober to use `credentials/insecure` instead.

To make this cleaner, I've opened this PR to add a config option to the grpc probe config to use plaintext auth.

### CODE REVIEW

**Comment by abaratif** [comment]
@manugarg Could you take a quick look here and let me know if this makes sense? Thx in advance for reviewing!

**Comment by manugarg** [comment]
Hey @abaratif, thanks for the PR! I think such an option makes sense. A couple of quick suggestions:

1. Can you please use the protoc version in the tools script to avoid the large diff in all the files.
2. For the config option itself, I think the name `allow_insecure` will probable make more sense. You can explain it in the comments.

P.S.: Is it for DoorDash? Good to know that DoorDash is still using Cloudprober :)

File: probes/grpc/grpc.go [hunk 1]
```diff
@@ -128,8 +129,11 @@ func (p *Probe) setupDialOpts() error {
...
 		p.dialOpts = append(p.dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(tlsCfg)))
 	}
+	if p.c.GetAllowInsecure() {
+		p.dialOpts = append(p.dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	}
 
-	if oauthCfg == nil && altsCfg == nil && p.c.GetTlsConfig() == nil {
+	if oauthCfg == nil && altsCfg == nil && p.c.GetTlsConfig() == nil && !p.c.GetAllowInsecure() {
 		p.dialOpts = append(p.dialOpts, grpc.WithTransportCredentials(local.NewCredentials()))
 	}
```
manugarg, line 138: I've a suggestion to improve the readability of this section. I think we can move transport credentials configuration to its own function like this:

```
func (p *Probe) transportCredOption() (grpc.DialOption, error) {
   if p.c.GetTlsConfig() != nil {
     return ...
   }
   ...
  ...
}

func (p *Probe) setupDialOpts() error {
   ...
   transportCredOpt, err := p.transportCredOption()
   if err != nil {
     return .. , err
   }
   p.dialOpts = append(p.dialOpts, transportCredOpt)
   ...
}
```

What do you think? Encapsulating transportCredentials logic in function will let us get rid of the if-else logic.

manugarg (commented):
Thanks for making the changes, @abaratif!

I've just one more suggestion to improve the readability.

probes/grpc/grpc.go
abaratif: sounds good to me, I took a stab at it and it does indeed look a bit cleaner 👍

probes/grpc/grpc.go
```diff
@@ -98,18 +99,23 @@ type probeRunResult struct {
...
 
-func (p *Probe) setupDialOpts() error {
+/*
+getTransportCreds attempts to fetch the transport creds from the probe config,
+returning any errors encountered
+if no creds are set, it returns local creds by default
+*/
+func (p *Probe) getTransportCreds() (grpc.DialOption, error) {
 	oauthCfg := p.c.GetOauthConfig()
 	if oauthCfg != nil {
```
manugarg, line 109: So, I'd leave the OAuth section under setupDialOpts itself (basically where it was), as it addresses RPC/service level authentication instead of transport level security.

abaratif: ah yes of course. I moved this back out. I also changed the behavior to have the caller be responsible for setting the default (local) option, so that we can keep the check similar to `if oauthCfg == nil && altsCfg == nil && p.c.GetTlsConfig() == nil `

probes/grpc/proto/config.proto
```diff
@@ -37,6 +37,9 @@ message ProbeConf {
   // Note that only one of ALTSConfig and TLSConfig can be enabled at a time.
   optional tlsconfig.TLSConfig tls_config = 9;
 
+  // if allow_insecure is set to true, TLS will not be used.
+  optional bool allow_insecure = 12;
```
manugarg, line 41: @abaratif my apologies, but thinking more about it, I think it will be more clear if we call this option "insecure_transport", so that it's more clear that we are using insecure transport.. we can still have OAuth if needed.

probes/grpc/grpc.go
```diff
@@ -98,38 +99,58 @@ type probeRunResult struct {
...
-		if err != nil {
-			return err
-		}
-		p.dialOpts = append(p.dialOpts, grpc.WithPerRPCCredentials(grpcoauth.TokenSource{TokenSource: oauthTS}))
-	}
-
+/*
+getTransportCreds attempts to fetch the transport creds from the probe config,
+returning any errors encountered
+if no creds are set, it nil
```
manugarg, line 105: This sentence seems incomplete.

manugarg (commented):
Sorry for requesting the option name change once again. Rest of change looks good to me now.

probes/grpc/grpc.go
abaratif: typo - fixed! ty

probes/grpc/proto/config.proto
abaratif: I agree it's more clear. updated now

manugarg (approved):

manugarg:
@abaratif so the change has been committed, but since I cut a release only yesterday, I'll wait for a couple of weeks before the next release. Can you use the "main" docker tag (or [download unreleased binaries](https://github.com/cloudprober/cloudprober/wiki/Download-unreleased-binaries)) in the meantime.

PR #273, opened 2023-02-09
[probes.grpc] support setting headers in probe config

abaratif:
This PR adds support for setting headers in grpc probe config.
Example:
```probe {
  name: "grpc_probe/test"
  type: GRPC
  targets {
    host_names: "localhost:8080"
  }
  grpc_probe {
    insecure_transport: true
    method: HEALTH_CHECK
    headers {
        name: "x-trace-id"
        value: "PDX01"
    }
    headers {
        name: "Cookie"
        value: "soft=chewy"
    }
  }

  interval_msec: 2000 # 2s
  timeout_msec: 1000  # 1s
}```

### CODE REVIEW

File: probes/grpc/grpc.go [hunk 1]
```diff
@@ -381,10 +382,24 @@ func (p *Probe) newResult(tgt string) *probeRunResult {
...
+	}
+	// create metadata from headers & attach to context
+	return metadata.NewOutgoingContext(ctx, metadata.New(parsed))
+}
+
 // Start starts and runs the probe indefinitely.
 func (p *Probe) Start(ctx context.Context, dataChan chan *metrics.EventMetrics) {
 	p.results = make(map[string]*probeRunResult)
-	p.updateTargetsAndStartProbes(ctx)
+	p.updateTargetsAndStartProbes(p.setHeaders(ctx))
```
abaratif, line 402: does this seem like the right place to touch the context?

abaratif:
@manugarg I've got another quick one for you, let me know what you think

probes/grpc/grpc.go
manugarg: Since it's a pretty quick operation, it will probably make more sense to do it here:
https://github.com/cloudprober/cloudprober/blob/fc9118e3f1784cfc1c90948a008b872048da7b7e/probes/grpc/grpc.go#L320

Otherwise we'll be just overloading a context that's being used for multiple purposes.

probes/grpc/grpc.go
```diff
@@ -381,10 +382,24 @@ func (p *Probe) newResult(tgt string) *probeRunResult {
 	}
 }
 
+// setHeaders attaches a list of headers to the given context
+// it iterates over the headers defined in the probe configuration
+func (p *Probe) setHeaders(ctx context.Context) context.Context {
```
manugarg, line 387: How about we call it ctxWithHeaders?
```suggestion
func (p *Probe) ctxWithHeaders(ctx context.Context) context.Context {
```

abaratif: sounds good - renamed

manugarg (approved):

abaratif:
@manugarg tysm for the merge & quick reviews here!!

manugarg:
@abaratif No problem! Thanks for the contribution.

PR #274, opened 2023-02-11
Adding new cloudprober surfacer for bigquery insertion.

Pulkit0110:
Adding a basic new bigquery surfacer which will insert metrics in the bigquery table. To make it more efficient, we're using batch insertion with a batch size of 1000 and a default interval of 10 seconds(it is configurable and can be changed while defining surfacer). 

It can take arguments like project name, bigquery dataset, bigquery table name, table columns name and type, batch insertion interval and bigquery timeout. At this point, it only supports data type string, int/integer, float/double and timestamp. For timestamp, the user will have to pass epoch time (in milliseconds). In future, we can add support for more data types.


kushal288:
@manugarg This is a PR adding a BQ surfacer that we were mentioning earlier. Please have a look and let us know for any feedback.

manugarg:
@Pulkit0110 @kushal288 This PR has a lot of unrelated files. Can you please clean it up?

Pulkit0110:
> @Pulkit0110 @kushal288 This PR has a lot of unrelated files. Can you please clean it up?

Its done.

surfacers/surfacers.go
```diff
@@ -158,6 +159,9 @@ func inferType(s *surfacerpb.SurfacerDef) surfacerpb.Type {
 		return surfacerpb.Type_DATADOG
 	case *surfacerpb.SurfacerDef_ProbestatusSurfacer:
 		return surfacerpb.Type_PROBESTATUS
+	case *surfacerpb.SurfacerDef_BigquerySurfacer:
```
kushal288, line 162: We need to add test for this

manugarg:
@Pulkit0110 Thanks for fixing the PR. I just wanted to say that I've not forgotten about this PR. I'll review it soon.

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,32 @@
...
+
+  optional int64 metrics_buffer_size = 4 [default = 100000];
+
+  repeated BQColumn columns = 5;
+
+  // This denotes the time interval after which data will be inserted in
+  // bigquery. Default is 10 seconds. So after every 10 seconds all the em in
+  // current will be inserted in bigquery in a batch size of 1000
+  optional int64 batch_insertion_time = 6 [default = 10];
+
```
manugarg, line 22: I think it will be more obvious to call it interval instead. Also, we include unit in the most config options.

```suggestion
  optional int64 batch_insertion_interval_sec = 6 [default = 10];

```

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,32 @@
...
+  // It represents bigquery client timeout in seconds. So, if bigquery insertion
+  // is not completed within this time period then the request will fail and the
+  // failed rows will be retried later.
+  optional int64 bigquery_timeout = 7 [default = 30];
+}
```
manugarg, line 27:
```suggestion
  optional int64 bigquery_timeout_sec = 7 [default = 30];
}
```

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,32 @@
...
+  optional string project_name = 1;
+
+  optional string bigquery_dataset = 2;
+
+  optional string bigquery_table = 3;
+
+  optional int64 metrics_buffer_size = 4 [default = 100000];
+
+  repeated BQColumn columns = 5;
+
```
manugarg, line 17: Let's add a comment on how this field works. Also, field name is not consistent. Above fields have bigquery_ prefix, columns doesn't. Is it on purpose?

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+			if err := inserter.Put(bqctx, results); err != nil {
+				s.l.Errorf("failed uploading probe results to Bigquery due to surfacer: %v", err)
+			}
+		}
+	}
+}
+
+func (s *Surfacer) writeToBQ(ctx context.Context, inserter iInserter) {
+	bigqueryInsertionTime := s.c.GetBatchInsertionTime()
+	ticker := time.NewTicker(time.Duration(bigqueryInsertionTime) * time.Second)
```
manugarg, line 148: Let's add a defer statement to stop this ticket:
```suggestion
	ticker := time.NewTicker(time.Duration(bigqueryInsertionTime) * time.Second)
	defer ticket.Stop()
```

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		bqRowMap[colName] = val
+	}
+	return bqRowMap, nil
+}
+
+func (s *Surfacer) batchInsertRowsToBQ(ctx context.Context, inserter iInserter) {
+	chanLen := len(s.writeChan)
+	bigqueryTimeout := time.Duration(s.c.GetBigqueryTimeout()) * time.Second
+	bqctx := context.Background()
```
manugarg, line 118: I think you should use the context provided in the args, no?

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+	bqctx, cancel := context.WithTimeout(bqctx, bigqueryTimeout)
+	defer cancel()
+
+	for i := 0; i < chanLen; i += batchSize {
+		var results []*bqrow
```
manugarg, line 123: I found the "results" (and later "result") name to be a bit confusing, will "row" not be better?

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+			em := <-s.writeChan
+
+			bqRowMap, err := s.parseBQCols(em)
+
+			if err == nil {
+				result := &bqrow{value: bqRowMap}
+				results = append(results, result)
+			} else {
+				s.l.Errorf("%v", err)
+			}
```
manugarg, line 135: I'd suggest swapping it around to make it like this:
```suggestion
			if err != nil {
			      s.l.Errorf("%v", err)
			      continue
			 }
			 results = append(results, &bqrow{value: bqRowMap})
```

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+		return val, nil
+	case "float", "double":
+		val, err := strconv.ParseFloat(label, 64)
+		if err != nil {
+			return nil, err
+		}
+		return val, nil
+	case "timestamp":
+		milliSec, err := strconv.ParseInt(label, 10, 64)
+		if err != nil {
```
manugarg, line 92: Metric's timestamp is not stored in the label, so if that's what you're looking for, you will need to get em.Timestamp.

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,183 @@
...
+
+	client, err := bigquery.NewClient(ctx, s.c.GetProjectName())
+	if err != nil {
+		s.l.Errorf("bigquery client can't be created: %v", err)
+		return err
+	}
+	defer client.Close()
+
+	inserter := client.Dataset(s.c.GetBigqueryDataset()).Table(s.c.GetBigqueryTable()).Inserter()
+	if inserter == nil {
```
manugarg, line 172: Should you return an error if inserter is nil?

manugarg (commented):
Hey @Pulkit0110,

Thanks again for adding this surfacer. Please take a look at the inline comments.

surfacers/bigquery/bigquery.go
Pulkit0110: It's not the metrics timestamp, here timestamp denotes the bigquery column type.

surfacers/bigquery/bigquery.go
Pulkit0110: Using only results now.

manugarg (commented):
I think something is missing in this surfacer. I don't see you making use of the metrics values. Just to make sure we're on the same page, here is the EventMetrics code:
https://github.com/cloudprober/cloudprober/blob/master/metrics/eventmetrics.go

Also, this code may benefit from looking at the postgres surfacer:
https://github.com/cloudprober/cloudprober/tree/master/surfacers/postgres

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,37 @@
...
+  optional string bigquery_table = 3;
+
+  optional int64 metrics_buffer_size = 4 [default = 100000];
+
+  // It represents the bigquery table columns.
+  //  bigquery_columns {
+  //    name: "id",
+  //    type: "string",
+  //  }
```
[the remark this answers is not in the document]
Pulkit0110: Here, bigquery_columns represent column name and type of bigquery table where data will be inserted. Using these columns, we're fetching all these column values from event metrics (as currently being done in the code). Since, all these values will be strings and the bigquery table may have different column type like integer, float, timestamp, etc. we need the column type so that string values can be parsed into valid data type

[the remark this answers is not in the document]
manugarg: Problem is I don't see you using EventMetrics values in the code anywhere. I see you calling only em.Label(..). Code as it is seems broken to me. Have you tried running it?

surfacers/bigquery/bigquery.go
Pulkit0110: If anyone wants to have their separate column of type timestamp which is defined in labels, this will be used for that case.

surfacers/bigquery/proto/config.proto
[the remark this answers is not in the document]
Pulkit0110: I've added metric name and value if that's what you are talking about. 

Yeah, it was working fine. I thought that all the column name and values will be passed through labels and therefore, was ignoring metrics values

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,45 @@
...
+  // is not completed within this time period then the request will fail and the
+  // failed rows will be retried later.
+  optional int64 bigquery_timeout_sec = 7 [default = 30];
+
+  optional int64 bigquery_batch_size = 8 [default = 1000];
+
+  // It will insert metric_name and value in columns 'metric_name' and
+  // 'metric_values' if set to true
+  optional bool insert_metric_values = 9 [default = true];
+}
```
manugarg, line 39: I don't understand why would someone not want to put the metric values in the data. Labels without values don't serve any purpose.

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,45 @@
...
+option go_package = "github.com/cloudprober/cloudprober/surfacers/bigquery/proto";
+
+message SurfacerConf {
+  optional string project_name = 1;
+
+  optional string bigquery_dataset = 2;
+
+  optional string bigquery_table = 3;
+
+  optional int64 metrics_buffer_size = 4 [default = 100000];
```
manugarg, line 14: Can you please group bigquery specific fields together and keep others after that:

project
bigquery_dataset
bigquery_table
bigquery_columns
bigquery_timeout

metrics_batch_size
batch_timer_sec

and so on..

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,45 @@
...
+  //    column_name: "id",
+  //    column_type: "string",
+  //  }
+  repeated BQColumn bigquery_columns = 5;
+
+  // This denotes the time interval after which data will be inserted in
+  // bigquery. Default is 10 seconds. So after every 10 seconds all the em in
+  // current will be inserted in bigquery in a batch size of 1000
+  optional int64 batch_insertion_interval_sec = 6 [default = 10];
+
```
manugarg, line 28: Can you please rename this to `batch_timer_sec` for consistency sake. That's what cloudwatch and stackdriver use.

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,45 @@
...
+  // It represents bigquery client timeout in seconds. So, if bigquery insertion
+  // is not completed within this time period then the request will fail and the
+  // failed rows will be retried later.
+  optional int64 bigquery_timeout_sec = 7 [default = 30];
+
+  optional int64 bigquery_batch_size = 8 [default = 1000];
+
```
manugarg, line 35: Similar to the above comment.. just call it metrics_batch_size.

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,45 @@
...
+  // It will insert metric_name and value in columns 'metric_name' and
+  // 'metric_values' if set to true
+  optional bool insert_metric_values = 9 [default = true];
+}
+
+message BQColumn {
+  optional string label = 1;
+  optional string column_name = 2;
+  optional string column_type = 3;
```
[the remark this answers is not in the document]
kushal288: Hey @manugarg,

Sorry for the misunderstandings here. I believe context of our use case is missing here. I will try to add some details here to clear up confusions.

We are using cloudprober to probe against cloud storage using cloud SDKs, which internally sends HTTP/gRPC requests.
Apart from several cloudprober metrics generated by default, we also insert the probe results in bigquery which has columns like HTTP_response_code, payload_size, workload_name etc..

@Pulkit0110 did some experiments  and found that using surfacer for inserting these columns in BQ is better than synchronous inserts.

So, basically for metrics we have  2 usecases - 
1. Default Metrics generated by cloudprober
2.  Custom metrics generated by external probes.

In 2 here, metric label's value can be of any BQ type, and we would need them to convert to their respective types.

e.g. a metric would look like 
```result{op=WRITE,response_code=200,payload_bytes=1024,workload_name=WRD} SUCCESS```

So, for 1 with out of the box metrics, labels have String type on both sides, while for custom metrics generated in external probes, we can have different type of values.

To align with default implementation, i think we should keep String as default type, and override if type is specifically provided in config.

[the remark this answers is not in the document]
Comment (Pulkit0110) [reply surfacers/bigquery/proto/config.proto:45]: @manugarg I've updated the code with metric values as it is being done in postgres. Can you have a look at it?

File: surfacers/bigquery/proto/config.proto (hunk 12 above)
Comment (Pulkit0110) [reply surfacers/bigquery/proto/config.proto]: Its more like whether they want to store metric values in bigquery or not. One possibility is that they can pass bigquery column name and value in the labels.

File: surfacers/bigquery/bigquery.go [hunk 17]
```diff
@@ -0,0 +1,216 @@
...
+	return convertToJSON(labels)
+}
+
+func (s *Surfacer) parseBQCols(em *metrics.EventMetrics) (map[string]bigquery.Value, error) {
+	bqRowMap := make(map[string]bigquery.Value)
+
+	if s.c.GetInsertMetricValues() {
+		for _, k := range em.MetricsKeys() {
+			bqRowMap["metric_name"] = k
+			bqRowMap["metric_value"] = em.Metric(k)
```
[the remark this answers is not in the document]
Pulkit0110: Done

manugarg:
@Pulkit0110 I'll take a look at it over the weekend.

Pulkit0110:
@manugarg did you get a chance to look at the latest changes?

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,296 @@
...
+	if !slices.Contains(excludedColumns, "metric_name") {
+		mapCopy["metric_name"] = metricName
+	}
+	if !slices.Contains(excludedColumns, "metric_value") {
+		mapCopy["metric_value"] = value
+	}
+	if !slices.Contains(excludedColumns, "metric_time") {
+		mapCopy["metric_time"] = timestamp
+	}
+	return mapCopy
```
manugarg, line 138: Please see my comment on the config below as to why you may not need this.

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (s *Surfacer) writeToBQ(ctx context.Context, inserter iInserter) {
+	bigqueryInsertionTime := s.c.GetBatchTimerSec()
+	ticker := time.NewTicker(time.Duration(bigqueryInsertionTime) * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			s.l.Infof("Context canceled, stopping the surfacer write loop")
+			s.batchInsertRowsToBQ(ctx, inserter)
```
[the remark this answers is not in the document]
Pulkit0110: What if there are still metrics in the surfacer channel, all those metrics would be dropped if we just return, thoughts?

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,46 @@
...
+
+  // It represents bigquery client timeout in seconds. So, if bigquery insertion
+  // is not completed within this time period then the request will fail and the
+  // failed rows will be retried later.
+  optional int64 bigquery_timeout_sec = 5 [default = 30];
+
+
+  // Any column name passes here won't be inserted in bigquery like
+  // 'metric_name', 'metric_value' or 'metric_time'
+  repeated string exclude_metric_column = 6;
```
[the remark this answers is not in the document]
Pulkit0110: We're only looking to drop the default metric values like metric name, value and timestamp and not the whole metric. As per my understanding, it will drop the whole metric including labels and other information which we don't wanna happen. Please correct me if I'm wrong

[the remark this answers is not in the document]
manugarg: So if you look at the example that @kushal288 shared above, your metrics are going to look like the following:
```
result{op=WRITE,response_code=200,payload_bytes=1024,workload_name=WRD} SUCCESS
```

I am assuming such metrics are generated by external probes, i.e. external probe writes a line like this that, and cloudprober creates metrics from such a line. My understanding is that you'd like to keep these metrics .. here metric name will be result, metric value is success, and everything else is labels.

In that case you can just allow metrics with name "result". All other metrics won't show up .. including labels. You see Cloudprober deals in the currency of metrics. Just labels don't make sense from Cloudprober's metrics model's point of view, and I don't really want bigquery surfacer to become very different from other surfacers. Does that make sense?

Here is the link to the previous comment from kushal:
https://github.com/cloudprober/cloudprober/pull/274#discussion_r1127776402

Ideally it should not be hard what you're trying to achieve. It's pretty much the same model as Postgres surfacer, except of typed columns.

In fact Cloud Bigtable team does something similar. They don't export data to bigquery, they export it to stackdriver, but also use metrics generated by their external probe and ignore the default metrics for the most part.

surfacers/bigquery/bigquery.go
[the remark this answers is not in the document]
manugarg: That's okay. Usually when you cancel context, you want to bail out quickly. Losing a few metrics is okay in such a scenario.

surfacers/bigquery/bigquery.go
Pulkit0110: Changed it

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,296 @@
...
+
+	for _, metricName := range em.MetricsKeys() {
+		val := em.Metric(metricName)
+
+		// Map metric
+		if mapVal, ok := val.(*metrics.Map); ok {
+			for _, k := range mapVal.Keys() {
+				bqRowMap := updateLabels(labels, mapVal.MapName, k)
+				bqRowMap = updateMetricValues(bqRowMap, metricName, mapVal.GetKey(k), em.Timestamp, s.c.GetExcludeMetricColumn())
+				bqMetrics = append(bqMetrics, &bqrow{value: labels})
```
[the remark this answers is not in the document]
Pulkit0110: Fixed

surfacers/bigquery/bigquery.go
Pulkit0110: removed these conditions

surfacers/bigquery/proto/config.proto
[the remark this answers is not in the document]
Pulkit0110: Made these metric column names configurable and also added support for allowing metrics with name

Pulkit0110:
@manugarg I've resolved all the comments. Can you have a look and let me know if more changes are required?

Pulkit0110:
@manugarg , did you get a chance to take a look at the latest changes?

manugarg:
Thanks for recent changes, @Pulkit0110! I've started reviewing it. Give me a couple of more days to finish.

manugarg (commented):
Hi Pulkit, 

It looks much better. Thank you for all the changes.

I added a few more comments, but most of them are suggestions except for the metric_col_type one. I am not sure if that's what you want there. Even if you do, looks like default should be numeric, not string.

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,294 @@
...
+			for _, k := range mapVal.Keys() {
+				bqRowMap := copyMap(baseRow)
+				bqRowMap[mapVal.MapName] = k
+				bqRowMap = updateMetricValues(bqRowMap, metricName, mapVal.GetKey(k), em.Timestamp, s.c)
+				out = append(out, &bqrow{value: bqRowMap})
+			}
+			continue
+		}
+
+		bqRowMap := copyMap(baseRow)
```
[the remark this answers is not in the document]
Pulkit0110: Yes we'll still need copy because we don't want to update baseRow at any point. For string type also, we're adding metrics to the copy of baseRow.

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,294 @@
...
+			bqRowMap = updateMetricValues(bqRowMap, metricName, "1", em.Timestamp, s.c)
+			out = append(out, &bqrow{value: bqRowMap})
+			continue
+		}
+
+		convertedVal, err := convertToBqType(s.c.GetMetricValueColType(), val.String())
+		if err != nil {
+			s.l.Errorf("error occurred while parsing for metric value %v: %v", val, err)
+			continue
+		}
```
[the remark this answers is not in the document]
Pulkit0110: Since metric val could be of any type (string, int, float etc), it would be better to parse it to desired type as well. I agree, it will only allow conversion for specific types but if we don't perform this conversion then we'll restrict to use column type as string or numeric in the bigquery.

surfacers/bigquery/proto/config.proto
```diff
@@ -0,0 +1,47 @@
...
+  optional int64 batch_timer_sec = 7 [default = 10];
+
+  optional int64 metrics_batch_size = 8 [default = 1000];
+
+  // Column name for metrics name, value and timestamp
+  optional string metric_time_col_name = 9 [default = "metric_time"];
+  optional string metric_name_col_name = 10 [default = "metric_name"];
+  optional string metric_value_col_name = 11 [default = "metric_value"];
+  optional string metric_value_col_type = 12 [default = "string"];
+}
```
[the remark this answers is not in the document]
Pulkit0110: Should I remove it and only support a particular type like string/numeric?

surfacers/bigquery/bigquery.go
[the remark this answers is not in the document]
manugarg: So you're anyway converting string, map, dist metrics to have numeric value only. Only types left are numeric types. I'd just assume remaining to be of numeric value type.

surfacers/bigquery/proto/config.proto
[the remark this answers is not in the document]
manugarg: Yes, I think so.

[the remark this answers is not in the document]
Pulkit0110: Removed it

surfacers/bigquery/bigquery.go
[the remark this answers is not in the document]
Pulkit0110: Removed it

manugarg (approved):

surfacers/bigquery/bigquery.go
```diff
@@ -0,0 +1,297 @@
+// Package bigquery implements surfacer for bigquery insertion.
+package bigquery
+
+import (
+	"context"
```
manugarg, line 5: I think you need to import encoding/json here.

manugarg (commented):
@Pulkit0110,

LGTM. Found a missing import. Can you please try to build and run it once more after all these changes. Thanks!

surfacers/bigquery/bigquery.go
Pulkit0110: Thanks for catching this. Fix it and also verified everything's working fine.

Issue #266, opened 2023-02-03
Make it easy to use k8s targets


manugarg:
Concept of resource discovery service (RDS) may be a bit hard to understand and make sense of. I think we should provide an easy way to use popular targets like k8s and file targets, we could do something like:

```
targets {
  spec: "k8s://services[/service_name]"
  selector {
    matchExpression {
      key: <label-key>
      value: <label-value>
      operator: ....
    }
  }
}
```

Basically we can just use kubernetes' label selector format: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement, and very easily convert this whole expression to an RDS client internally. We'll also provide a way to override RDS server options in the `targets` stanza itself. All in all, we'll end up adding following fields to targets definition:

* spec
* selector
* [ip_config](https://github.com/cloudprober/cloudprober/blob/d1b4071e8cc59764b087b56498552ae67287867e/rds/proto/rds.proto#L56) (to keep the ability to specify IP version)
* rds_server_options (to provide scalable solution where you may want to point to another RDS server)

All these options will be used to build RDS client internally. I guess the idea is to hide the RDS concept for simple Cloudprober setups (startups) -- users can still use RDS concepts (e.g. a centralized server) for large installations (like Google -- in large setups it helps to centralize targets discovery).

manugarg:
I think we should also allow filtering by namespace and port. So something like:

```
targets {
  spec: "k8s://services[/service_name]"
  labelSelector {
    matchExpression {
      key: <label-key>
      value: <label-value>
      operator: ....
    }
  }
  namespace: (regex)
  ports: (regex)
}
```

manugarg:
I am not sure about the above format. I think following may work better:
```
targets {
  k8s {
    services: ""
    namespace: "dev"
    labelSelector: "k8s-app"
    labelSelector: "role=frontend"
  }
}
```

It's yaml equivalent will be:

```
targets:
  k8s:
    services:
    namespace: "dev"
    labelSelector:
    - "k8s-app"
    - "role=frontend"
```

This will select all the services in namespace "dev" that have the `k8s-app` label (any value) and a `role=frontend` label.

manugarg:
This is done. Example config:
https://github.com/cloudprober/cloudprober/pull/276#issuecomment-1432258535

PR #276, opened 2023-02-12
[targets.k8s] Native k8s targets support

manugarg:
Allow adding k8s targets without knowing about RDS.


manugarg:
This implements #266.

manugarg:
Example config:

```
probe {
    type: HTTP
    name: "http_endpoints"
    targets {
        # All endpoints that don't contain app=cloudprober label
        k8s {
            endpoints: ""
            labelSelector: "app!=cloudprober"
        }
    }
    interval: "10s"
    timeout: "5s"

    http_probe {}

    validator {
        name: "status_code_2xx"
        http_validator {
            success_status_codes: "200-499"
        }
    }
}

probe {
    type: DNS
    name: "dns_endpoints"
    targets {
        # All endpoints ending in -dns in kube-system namespace
        k8s {
            endpoints: ".*-dns"
            namespace: "kube-system"
        }
    }
    interval: "10s"
    timeout: "5s"

    dns_probe {
        resolved_domain: "www.google.com"
        min_answers: 1
        query_type: A
    }
}
```

manugarg:
I think looking pretty good now. We just need to add some tests.

zmailloux:
Hey @manugarg, if I'm looking at the code correctly, by default 
`        # All endpoints ending in -dns in kube-system namespace
        k8s {
            endpoints: "-dns"
            namespace: "kube-system"
        }`
Would actually include everything with "-dns" in it, whereas one would expect that ".*-dns" would be needed. This at least seems to be the case in our scenario where we are filtering on `endpoints: "server"` and it picks up endpoints with names like "dynamo-server" as well. I guess I would have expected having to define ".*server" in order to see that returned.

zmailloux:
That makes sense! After I had wrote it I had removed it as I figured that it's intended to act as such. Makes sense, and thanks for clarifying!

Issue #191, opened 2022-09-27
[common.oauth] Support JSON format from Bearer token sources


manugarg:
See this: https://github.com/cloudprober/cloudprober/blob/master/common/oauth/proto/config.proto

Currently bearer token type expects file and command sources to return only access token string. We should support reading JSON as well, so that we can get additional information about the token, e.g. expiration time (typically: expires_in).

manugarg:
Note #275 reduces the need for this, but we can still do this.

PR #278, opened 2023-02-13
[oauth] Let bearer token type handle JSON tokens.

manugarg:
Implements #191.


manugarg:
This PR keeps the old behavior working, while implementing new behavior for sources that return token in JSON format.

PR #282, opened 2023-02-20
Increase buffer size to 1000 and add a timer for the cloudwatch surfacer

robpickerill:
I'd missed this originally but around August 2022 the limit for max metrics in a putmetricdata call was increased from 20 to 1000. We should implement this to save costs on the putmetricdata calls for any AWS users here.

Additionally, with the increased buffer size I opt'ed to add a timer for the maximum amount of time to keep metrics within the buffer before publishing them to Cloudwatch.


surfacers/cloudwatch/cloudwatch.go
```diff
@@ -50,7 +47,45 @@ type CWSurfacer struct {
 
 	// A cache of []types.MetricDatum's, used for batch writing to the
 	// cloudwatch api.
-	cwMetricDatumCache []types.MetricDatum
+	metricDatumCache       []types.MetricDatum
+	metricDatumCacheTicker *time.Ticker
```
manugarg, line 51: Instead of having ticker be part of this struct, I think it will be better to move creation and stopping of ticker to processIncomingMetrics function. That way we won't leak a goroutine if at some some point we make surfacers similar probes, i.e. stoppable/start-able through some API/

surfacers/cloudwatch/cloudwatch.go
```diff
@@ -50,7 +47,45 @@ type CWSurfacer struct {
...
+// cloudwatch surfacer to receive EventMetrics over the buffered channel.
+func (cw *CWSurfacer) Write(ctx context.Context, em *metrics.EventMetrics) {
+	select {
+	case cw.writeChan <- em:
+	default:
+		cw.l.Error("Surfacer's write channel is full, dropping new data.")
+	}
 }
 
 func (cw *CWSurfacer) processIncomingMetrics(ctx context.Context) {
```
manugarg, line 91: Regarding my comment above about moving ticker creation here:
```suggestion
func (cw *CWSurfacer) processIncomingMetrics(ctx context.Context) {
    batchTimeTicker := time.NewTicker(...)
    defer batchTimeTicker.Stop()
```

tools/gen_pb_go.sh
```diff
@@ -56,6 +56,12 @@ else
 fi
 
 arch=$(uname -m)
+
+# protoc for arm64 on darwin is not available. Use universal binary instead.
+if [ "${os}" == "osx" ] && [ "${arch}" == "arm64" ]; then
+  arch="universal_binary"
+fi
+
```
manugarg, line 64: Curious, why universal binary? :)

Good catch regardless though. We need to change the value of arch for MacOS/ARM64 as protoc seems to be uploading binaries with the aarch64 extension.
https://github.com/protocolbuffers/protobuf/releases/tag/v21.10

manugarg (commented):
A couple of minor comments, looks good to me otherwise.

surfacers/cloudwatch/cloudwatch.go
robpickerill: Ah yeah good call, and I did write it like this at one point, but opt'ed to add it to the struct as I wanted to control the reset when publishing a full batch. Let me rework a little.

tools/gen_pb_go.sh
robpickerill: Honestly, I just blindly never saw the aarch64 binaries, let me update this.

manugarg (approved):
One minor comment regarding the default for the timer.

surfacers/cloudwatch/proto/config.proto
```diff
@@ -21,4 +21,17 @@ message SurfacerConf {
...
+  // limit is reached. 1000 is the maximum number of metrics
+  // supported by the Cloudwatch PutMetricData API.
+  // Metrics will be published when the timer expires, or the buffer is
+  // full, whichever happens first.
+  optional int32 metrics_batch_size = 4 [default = 1000]; 
+
+  // The maximum amount of time to hold metrics in the buffer (above).
+  // Metrics will be published when the timer expires, or the buffer is
+  // full, whichever happens first. 
+  optional int32 metrics_publish_timer_sec = 5 [default = 60];
```
manugarg, line 36: One more thing, regarding this field's name. I am not really sure what we should use, but for stackdriver we use batch_timer_sec:
https://github.com/cloudprober/cloudprober/blob/77a13fc548b7fe0fead31523a5745727e608b124/surfacers/stackdriver/proto/config.proto#L13

What do you think about using the same name?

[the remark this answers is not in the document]
robpickerill: yeah, 30s is good, updated

robpickerill: yeah, I think consistency wins here, renamed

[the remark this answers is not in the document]
manugarg: Hey @robpickerill, I think you missed including this change. I still see 60.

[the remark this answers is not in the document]
robpickerill: my apologies, fixed and rebased the duplicate commit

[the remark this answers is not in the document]
robpickerill: I also updated the docs, I think that should be good to go now, apologies for the oversight on the timer change.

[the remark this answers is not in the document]
manugarg: No problem at all! :) Thank you!

Issue #288, opened 2023-02-26
[probes.http] Add an option to disable redirect follow.


manugarg:
Go's HTTP client follows redirects by default. We should provide an option to disable that behavior.

PR #289, opened 2023-02-26
add max redirects

robpickerill:
Adds a max redirect config option, with the default set to the [go client default](https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/client.go;l=802-807;drc=e7201cfa9fe638053b7c269c23bf408223600bc9;bpv=1;bpt=1).

This is missing tests, but if this if you were thinking along these lines for https://github.com/cloudprober/cloudprober/issues/288 then I'll continue and add tests.


probes/http/proto/config.proto
```diff
@@ -90,6 +90,10 @@ message ProbeConf {
   // Maximum idle connections to keep alive
   optional int32 max_idle_conns = 17 [default = 256];
 
+  // The maximum amount of redirects the HTTP client will follow. 
+  // To disable redirects, use max_redirects = 0.
+  optional int32 max_redirects = 18 [default = 10];
```
manugarg, line 95: I think for default we'll leave the default behavior.. whatever Go's net/http client does.

So set client's CheckRedirect only if p.c.MaxRedirects != nil.

probes/http/http.go
```diff
@@ -474,6 +474,14 @@ func (p *Probe) clientsForTarget(target endpoint.Endpoint) []*http.Client {
...
 			clients[i] = &http.Client{Transport: p.baseTransport}
 		}
+
+		clients[i].CheckRedirect = func(req *http.Request, via []*http.Request) error {
+			if len(via) >= int(p.c.GetMaxRedirects()) {
+				return http.ErrUseLastResponse
+			}
+
+			return nil
+		}
```
manugarg, line 484: One optimization we can likely do is to define checkRedirect function once as part of the probe struct and here just point client.CheckRedirect to that function.

Something like in Init():

```
if p.c.MaxRedirect != nil {
  p.checkRedirect = func(...) {
  }
}
```

manugarg (commented):
Thanks for the PR, Rob!

Added a couple of suggestions.

robpickerill:
Thanks Manu! I cleaned this up and its ready for you to take another look

probes/http/proto/config.proto
robpickerill: Yeah I'd thought about this after I'd pushed - if the go client behaviour changes we won't follow it, good call

probes/http/http_test.go
```diff
@@ -701,3 +699,56 @@ func TestGetTransport(t *testing.T) {
...
+
+func TestProbeInitRedirectsNotSet(t *testing.T) {
+	p := &Probe{}
+
+	opts := &options.Options{
+		Targets:  targets.StaticTargets("test.com"),
+		Interval: 10 * time.Millisecond,
+		ProbeConf: &configpb.ProbeConf{
+			MaxRedirects: nil,
+		},
```
manugarg, line 743: I'll probably just set it to default config, to verify the default behavior:

```suggestion
		ProbeConf: &configpb.ProbeConf{}
```

PR #302, opened 2023-02-28
[targets.k8s] Automatically add anchors if not already configured

manugarg:


manugarg:
See https://github.com/cloudprober/cloudprober/pull/276 for the background.

PR #305, opened 2023-03-02
[surfacer.postgres] Switch to newer pgx postgres driver

manugarg:


markoposavec:
Hey!
Is there a plan of merging this?
pgx postgres driver allows setting multiple hosts in connection string which would solve some issues for us 😄

manugarg:
Sorry for leaving it without a comment. I didn't merge this because new driver doubles the size of the binary. I'll take another look at it.

manugarg:
> Sorry for leaving it without a comment. I didn't merge this because new driver doubles the size of the binary. I'll take another look at it.

Binary size increase is only about 8% now. Given the benefits of pgx driver over lib/pq, I'll try merge it soon.

Issue #307, opened 2023-03-09
Issues with TLS for GRPC Probe


zmailloux:
**Describe the bug**
Hello Manu! I'm trying to use the GRPC probe type and had been running into problems and ultimately came up with a workaround, but I was hoping to revisit it and see if you had any ideas. I have a the following two gRPC health probes, one uses an external CLI I have loaded onto the container, and another uses the GRPC probe type within cloudprober. In the example below, "grpc probe v1" works, and "grpc probe v2" does not. It has been like this for quite some time now, and I've verified that they are hitting the same address by the log message.

**Cloudprober Version**
1.25.0

**To Reproduce**
```probe {
    type: EXTERNAL
    name: "grpc probe v1"
    targets {   
        k8s {
            endpoints: "server"
            namespace: "sample"
            portFilter: "service-discovery-ssl"
        }
    }

    interval: "30s"
    timeout: "5s"

    latency_unit: "s"
    latency_metric_name: "latency_dist"

    latency_distribution {
      explicit_buckets: "0.01,0.1,0.15,0.2,0.25,0.35,0.5,0.75,1.0,1.5,2.0,3.0,4.0,5.0,10.0,15.0,20.0"
    }

    external_probe {
      command: "grpc_health_probe -addr=@address@:@port@ -tls -tls-ca-cert=/shared/pki/root_ca.pem -tls-client-cert=/shared/pki/leaf_intermediate_chain.pem -tls-client-key=/shared/pki/private_key.pem -tls-server-name=global.identity"
    }
}

probe {
    type: GRPC
    name: "grpc probe v2"
    targets {
        k8s {
            endpoints: "server"
            namespace: "sample-app"
            portFilter: "service-discovery-ssl"
        }
    }
    interval: "10s"
    timeout: "5s"

    grpc_probe {
        method: HEALTH_CHECK
        tls_config {
          ca_cert_file: "/shared/pki/root_ca.pem",
          tls_cert_file: "/shared/pki/leaf_intermediate_chain.pem",
          tls_key_file: "/shared/pki/private_key.pem",
          server_name: "global.identity"
        }
        health_check_ignore_status: false
    }

    latency_unit: "s"
    latency_metric_name: "latency_dist"

    latency_distribution {
      explicit_buckets: "0.01,0.1,0.15,0.2,0.25,0.35,0.5,0.75,1.0,1.5,2.0,3.0,4.0,5.0,10.0,15.0,20.0"
    }
}
[...]

PR #309, opened 2023-03-10
[probes.grpc] Use discovered target's IP for connection

manugarg:


manugarg:
Solves the bug reported in #307.

PR #357, opened 2023-04-21
add cache and timer for dd

robpickerill:
Implementing a configurable cache and timer for Datadog, very similar to the Cloudwatch implementation.

PR #358, opened 2023-04-21
add compression when posting metrics to dd

robpickerill:
Adds gzip compression when posting metrics to Datadog.


robpickerill:
I think adding a configuration option for compression maybe the best way to go here, will rework the PR to support that.

surfacers/datadog/proto/config.proto
```diff
@@ -17,4 +17,7 @@ message SurfacerConf {
 
   // Datadog server, default: "api.datadoghq.com"
   optional string server = 4;
+
+  // Compress metrics before sending them to Datadog.
+  optional bool compress = 7 [default = true];
```
robpickerill, line 22: Pending #357

surfacers/datadog/client.go
```diff
@@ -119,3 +130,19 @@ func (c *ddClient) submitMetrics(ctx context.Context, series []ddSeries) error {
 
 	return nil
 }
+
+func compressPayload(b []byte) (*bytes.Buffer, error) {
+	var buf bytes.Buffer
+	gz, err := gzip.NewWriterLevel(&buf, gzip.BestCompression)
```
manugarg, line 136: Can we use this function:
https://github.com/cloudprober/cloudprober/blob/9ecd8a25aeb66c289d8b21f052d36ab89fbc9bf2/surfacers/common/compress/compress.go#L99

One diff is base64 conversion, in the Compress function, gzipped content is converted to base64. I think we don't need that here? If that's the case let's keep it as it is.

manugarg (commented):
Looks good. Just one minor suggestion.

surfacers/datadog/client.go
robpickerill: Yeah I see, I didn't spot that function and it's a little duplication here. 

But I don't think the base64 encoding is required, but before we commit, let me try it with Datadog and come back to you.

robpickerill: The base64 encoding was a problem when testing with Datadog, and I couldn't see a way to inform dd that the encoding type was base64 encoded and compressed. I opt'ed for the private compressPayload function in the end if your ok with that?

surfacers/datadog/proto/config.proto
robpickerill: I rewrote the config to drop the overwritten default value, and I refactored the PR a little to remove the double negative evaluations in the logic.

robpickerill:
I was also thinking about adding a minimum size for compression, like 1kb or 2kb, let me know if you think that would be preferable here.

surfacers/datadog/client.go
manugarg: Yes. Thanks for checking!

manugarg (approved):
Thanks, Rob!

PR #379, opened 2023-05-10
[probes.external] Add support for environment variables

manugarg:


manugarg:
Thanks to @alex-laties for the idea in #205.

PR #395, opened 2023-05-26
Synchronize access to cmdRunning

apolcyn:
I haven't actually ran into any issue related to this, but from my reading it looks like access to `cmdRunning` should be synchronized

### DIFF
--- a/probes/external/external.go
+++ b/probes/external/external.go
@@ -88,6 +88,7 @@ type Probe struct {
 	labelKeys  map[string]bool // Labels for substitution
 	requestID  int32
 	cmdRunning bool
+	cmdRunningMu sync.Mutex // synchronizes cmdRunning
 	cmdStdin   io.Writer
 	cmdStdout  io.ReadCloser
 	cmdStderr  io.ReadCloser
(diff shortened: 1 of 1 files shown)


manugarg:
Hey @apolcyn,

p.cmdRunning is accessed only from the startCmdIfNotRunning function:
https://github.com/cloudprober/cloudprober/blob/09fc42294a46f1f1bfb4c03482057b3c87fa96b5/probes/external/external.go#LL210C17-L210C37

And, this function itself is called only from the runServerProbe:
https://github.com/cloudprober/cloudprober/blob/09fc42294a46f1f1bfb4c03482057b3c87fa96b5/probes/external/external.go#L424

and, runServerProbe is called sequentially, i.e. no parallelism, at probe ticks.

Based on this, cmdRunning is never accessed concurrently.

Does it make sense?

apolcyn:
Hi @manugarg 

`startCmdIfNotRunning` spawns a goroutine which sets `p.cmdRunning = false`

So it looks like `startCmdIfNotRunning` can be accessed concurrently between `startCmdIfNotRunning` and that goroutine, leading to a possible data race. No?

manugarg:
@apolcyn, you're right. I missed that goroutine. Thanks for noticing it. I'll try to add a test that will involve the execution of that goroutine (We test for [race](https://github.com/cloudprober/cloudprober/blob/09fc42294a46f1f1bfb4c03482057b3c87fa96b5/.github/workflows/go.yml#L69) in the tests but they are only useful for the paths covered by tests).

manugarg:
Thank you so much again. I've added a test (#395) that would have caught it:
Failure:
https://github.com/cloudprober/cloudprober/actions/runs/5129660385/jobs/9227515240?pr=397

apolcyn:
Great, thanks for the test @manugarg

PR #396, opened 2023-05-29
[cloudprober] Add support for reading files from S3 and GCS.

manugarg:
This applies to all kinds of file resources:
 - File based targets
 - TLS certificate files
 - OAuth token files
 - Cloudprober config itself.


manugarg:
Note about this change: It will increase convenience but it will make cloudprober run dependent on external configs. For the most part it's okay -- it will only make debugging a bit harder as you'll have to find the external file content when cloudprober started running. Maybe, as @theduke suggested above, we can cache file contents locally, which will be useful in two cases:

a. to debug.
b. for the restarts where external file service is unavailable and you need to restart. Note, ideally, you don't want to start if you cannot get the latest content, so maybe we'll provide a flag to use cache if actual file is not available -- it will not be the default behavior.

I'll consider hashing this out and implementing it in a different pull request / issue.

PR #418, opened 2023-07-03
[probes.http] Include headers configured with `header` option in requests

nrhtr:
Hi,

I wanted to add HTTP headers to a probe configuration and I found the "header" option first, as well as a "headers" option which appeared to be the deprecated equivalent (if I'm understanding correctly).

Adding headers to the "header" option seemed to have no effect, it appears only "headers" is being checked. This should resolve that.

Hopefully this is correct. It seems to work for me :)

### DIFF
--- a/probes/http/request.go
+++ b/probes/http/request.go
@@ -128,6 +128,14 @@ func (p *Probe) httpRequestForTarget(target endpoint.Endpoint) *http.Request {
 		req.Header.Set(header.GetName(), header.GetValue())
 	}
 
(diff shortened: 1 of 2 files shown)


manugarg:
Thank for catching and fixing this, @nrhtr! Last change (https://github.com/cloudprober/cloudprober/pull/375) went through quite a few iterations and I missed that I had added the `header` option in config proto.

(Sorry for the delay in reviewing, I was on vacation.)

PR #428, opened 2023-07-31
[tls] Provide an option to reload TLS certificates.

manugarg:
Couple of things for the records:
* File cache is maintained by the file name. If multiple probes are using the same certificate, they will all access the same content, and they will all take the read lock, but read locks are cheap.
* Certificate reload may get triggered sporadically if multiple probes (or targets) use the same certificate but specify different reload interval. This is unlikely and quite harmless I think, so not trying to solve that.


common/tlsconfig/tlsconfig.go
```diff
@@ -19,11 +19,34 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("common/tlsconfig: error reading TLS cert file (%s): %v", certFile, err)
+	}
+
+	keyPEMBlock, err := readFile(keyFile)
+	if err != nil {
+		return nil, fmt.Errorf("common/tlsconfig: error reading TLS key file (%s): %v", keyFile, err)
+	}
+
+	cert, err := tls.X509KeyPair(certPEMBlock, keyPEMBlock)
```
cbroglie, line 46: There is a CPU cost involved in parsing the key pair on each load, even if disk read is skipped. What do you think about caching the parsed key pair instead?

common/tlsconfig/tlsconfig.go
```diff
@@ -44,20 +67,21 @@ func UpdateTLSConfig(tlsConfig *tls.Config, c *configpb.TLSConfig) error {
...
 		}
 
-		cert, err := tls.X509KeyPair(certPEMBlock, keyPEMBlock)
-		if err != nil {
-			return fmt.Errorf("common/tlsconfig: error initializing cert from cert key pair: %v", err)
+		if c.GetReloadIntervalSec() > 0 {
+			tlsConfig.GetCertificate = func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				return loadCert(c.GetTlsCertFile(), c.GetTlsKeyFile(), time.Duration(c.GetReloadIntervalSec())*time.Second)
+			}
+			return nil
```
cbroglie, line 81: This early return skips setting `ServerName`

manugarg: Really good catch. Thanks Chris! Fixing in #440. Also added tests to catch this in future.

common/tlsconfig/tlsconfig.go
manugarg: That's a good point. That means we'll have to keep a global cert cache too -- in that case it may be better to move cache refresh here itself. I'll look into this.

cbroglie: #441 LGTM. I can test it out the reloading this week.

PR #435, opened 2023-08-02
Add a health handler.

manugarg:
It gets activated after Cloudprober has fully started. It can be used for readiness and liveness probes.

### DIFF
--- a/cloudprober.go
+++ b/cloudprober.go
@@ -241,7 +241,8 @@ func Start(ctx context.Context) {
 	defer cloudProber.Unlock()
 
 	// Default servers
-	httpSrv := &http.Server{Handler: runconfig.DefaultHTTPServeMux()}
+	srvMux := runconfig.DefaultHTTPServeMux()
+	httpSrv := &http.Server{Handler: srvMux}
 	grpcSrv := runconfig.DefaultGRPCServer()
 
 	// Set up a goroutine to cleanup if context ends.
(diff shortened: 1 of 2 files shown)


cloudprober.go
```diff
@@ -271,6 +272,9 @@ func Start(ctx context.Context) {
 	}
 
 	cloudProber.prober.Start(ctx)
+	srvMux.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
```
ashi009, line 275: I think there is a race condition there: before `prober.Start()` is called, the `httpSrv` is already up, and all traffic to "/health" will be redirected to "/status" with a 302. Which may or may not be a problem. But I think we can do better here.

cloudprober.go
```diff
@@ -271,6 +272,9 @@ func Start(ctx context.Context) {
...
+		fmt.Fprintf(w, "OK")
+	})
```
ashi009, line 277: While we are here, should we also implement a health check service for grpc endpoint? i.e. https://pkg.go.dev/google.golang.org/grpc/health

cloudprober.go
manugarg: Yeah, I've been testing this and ran into the same problem. Ideally we should return 404 for non-root URLs instead of redirecting all of them. We should redirect only the absolute root. I'll update the PR soon.

ashi009 (approved):

Issue #423, opened 2023-07-26
"resp-code" metric of HTTP probe can change in surfacer


jinmeiib:
**Describe the bug**
I'm not sure if this is intentional or a bug, but this is the description:
The "resp-code" metric of an HTTP probe passed to a surfacer may be different from the metric at the time the event is created

**Cloudprober Version**
0.12.6, but the behavior seems to be the same on master as of commit 8056812.

**To Reproduce**
1. configure an http probe, e.g., with the following parameter:
```
  interval_msec: 2000  # 2s
  timeout_msec: 1000   # 1s
  stats_export_interval_msec: 2000
```
2.  write a custom surfacer that simply prints the event metrics with some delay:
```
func (s *mySurfacer) Write(ctx context.Context, em *metrics.EventMetrics) {
    time.Sleep(5 * time.Second)
    fmt.Printf("event metrics: %v\n", em)
}
```
3. Run the prober.  Then, on the first call on `Write`, you'll see something like:
```
event metrics: ...total=1 success=1 latency=712057.779 timeouts=0 resp-code=map:code,200:3
```
I'm afraid this is confusing, since it looked like there were 3 responses (of code 200) while the total in the metric is 1.
Is this an intentional behavior for some specific reason?

**Additional context**
I've not tested it, but from code inspection, the same possible issue can happen for "resp-body" and "validation_failure" metrics.

If this is not intentional, one obvious fix is to clone the resp code map to be added to the event metrics:
```
	em := metrics.NewEventMetrics(ts).
		AddMetric("total", metrics.NewInt(result.total)).
		AddMetric("success", metrics.NewInt(result.success)).
		AddMetric(p.opts.LatencyMetricName, result.latency).
		AddMetric("timeouts", metrics.NewInt(result.timeouts)).
		AddMetric("resp-code", result.respCodes.Clone()) // make a clone of respCodes
```

manugarg:
Thanks for the bug report @jinmeiib!

I am not sure what's going on here, but I am unable to reproduce this with the stock surfacers:
```
[samwise 15:51:51 ~/code/cloudprober(master)]$ go run ./cmd/. --config_file ~/cloudprober_423.cfg
cloudprober 1690411914404336026 1690411914 labels=ptype=http,probe=http,dst=www.google.com total=1 success=1 latency=227840.744 timeouts=0 resp-code=map:code,200:1
[...]

PR #437, opened 2023-08-03
[metrics] Clone latency and other similar metrics

manugarg:
One possible repercussion of this could be increased memory allocation and de-allocations for large number of probes/targets. Rough calculation for the record:

* Based on the map cloning [benchmark](https://github.com/cloudprober/cloudprober/blob/83add1a5e02d2714d71a10b7cb3ed42d8bf27009/metrics/map_test.go#L160), each map clone causes N+6 allocations where N is number of keys. Assuming each resp-code entry takes about 24 bytes (8 byte for keys array, 16 bytes for key-value map entry) and there can be say 10 resp-code in the map, and taking into account some static overhead, we're looking at around 300 bytes per resp-code map cloning.

* Latency distribution requires explicit configuration so it's less common. As per benchmark, distribution cloning requires fixed number of allocation (4), regardless of number of buckets (O(1)). Let's say there are 24 buckets. Each bucket takes 8 bytes for bucket definition (lower bounds array) and 8 bytes for counts. We're looking at about 400 bytes per latency distribution cloning.

Assuming there are 10000 targets (across all probes) exporting data every 10s, we're looking at about 700 kB/second of allocations. This is pretty much an extreme case though and all these allocations will be spread over.

### DIFF
--- a/probes/dns/dns.go
+++ b/probes/dns/dns.go
@@ -103,7 +103,7 @@ func (prr probeRunResult) Metrics() *metrics.EventMetrics {
 	return metrics.NewEventMetrics(time.Now()).
 		AddMetric("total", &prr.total).
 		AddMetric("success", &prr.success).
-		AddMetric(prr.latencyMetricName, prr.latency).
+		AddMetric(prr.latencyMetricName, prr.latency.Clone()).
 		AddMetric("timeouts", &prr.timeouts).
 		AddMetric("validation_failure", prr.validationFailure)
 }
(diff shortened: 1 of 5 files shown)


manugarg:
I am anyway working on improving the Map implementation (#443), so changes here are less of concern. This should fix #423.

manugarg:
https://github.com/cloudprober/cloudprober/pull/533 kind of implemented more strict cloning -- all EventMetrics are cloned before being put on the data channel. That changes eliminates the need of this PR.

PR #443, opened 2023-08-05
[metrics] Improve Map implementation.

manugarg:
Map implementation is complex and expensive primarily to deal with both int64 and float64 values. Map is used almost always as a counter (e.g. resp-code, validation failure map), except for external probe output parsing where we can encounter a float map.

To simplify the Map implementation, split it into two implementations: `Map` (for int64 counters) and `MapFloat` (for any float64 values map).


manugarg:
Merged through #444.

PR #476, opened 2023-08-23
Adding the stringSplit macro to the Go template component

xenbyte:
When utilizing Cloudprober's Go template component to generate configuration files, a scenario arose in which I required the capability to divide a pre-defined string into a slice of strings. Regrettably, the absence of a suitable macro impeded this process. As a solution, I have introduced the `stringSplit` macro, which seamlessly employs the `strings.Split()` function to accomplish this task.

### DIFF
--- a/config/config_tmpl.go
+++ b/config/config_tmpl.go
@@ -1,6 +1,7 @@
 // Copyright 2017-2023 The Cloudprober Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
 //      http://www.apache.org/licenses/LICENSE-2.0
--- a/config/config_tmpl.go
+++ b/config/config_tmpl.go
@@ -26,94 +26,112 @@ variable map (usually GCP metadata variables) and some predefined macros.
 
 Cloudprober configs support some macros to make configs construction easier:
 
-	env
-		Get the value of an environment variable.
+		env
+			Get the value of an environment variable.
 
-		Example:
+			Example:
+
+			probe {
+			  name: "dns_google_jp"
+			  type: DNS
+			  targets {
+			    host_names: "1.1.1.1"
+			  }
+			  dns_probe {
+			    resolved_domain: "{{env "TEST_DOM"}}"
+			  }
+			}
+
+			# Then run cloudprober as:
+			TEST_DOM=google.co.jp ./cloudprober --config_file=cloudprober.cfg
+
+		gceCustomMetadata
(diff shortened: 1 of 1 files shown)


config/config_tmpl.go
```diff
@@ -220,6 +239,21 @@ func ParseTemplate(config string, sysVars map[string]string, getGCECustomMetadat
...
+					slice := strings.Split(val, sepVal)
+					return slice, nil
+				} else {
+					return nil, errors.New("separator must be a string")
+				}
+			} else {
+				return nil, errors.New("input argument must be a string")
+			}
+		},
 	}
```
manugarg, line 257: Can you call it just splitList to make it consistent with helm templates?
Example: https://github.com/werf/werf/blob/3920241c1a463c843055d8a47e1b9efab150b251/docs/.helm/templates/20-ingress.yaml#L1

and, splitList reverses the argument list:
splitList sep string

config/config_tmpl.go
```diff
@@ -220,6 +239,21 @@ func ParseTemplate(config string, sysVars map[string]string, getGCECustomMetadat
...
+				if sepVal, sepOk := sep.(string); sepOk {
+					slice := strings.Split(val, sepVal)
+					return slice, nil
+				} else {
+					return nil, errors.New("separator must be a string")
+				}
+			} else {
+				return nil, errors.New("input argument must be a string")
+			}
+		},
```
manugarg, line 256: I'd keep it simple:
```suggestion
		// splitList splits the given string into a list.
		"splitList": func(sep, input string) []string { return strings.Split(input, sep) },
```

manugarg (commented):
Thanks for the PR, @xenbyte!

Please see the inline comments.

config/config_tmpl.go
```diff
@@ -26,94 +26,112 @@ variable map (usually GCP metadata variables) and some predefined macros.
...
-	env
-		Get the value of an environment variable.
+		env
+			Get the value of an environment variable.
+
+			Example:
+
+			probe {
+			  name: "dns_google_jp"
+			  type: DNS
```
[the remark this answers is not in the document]
xenbyte: Yeah it was a weird to me as well why my editor did that.
I'll see if I can fix that

xenbyte:
Hey @manugarg, just a heads up that I've removed the comments from the code document due to ongoing formatting glitches in my editor. It was causing some significant changes in the diffs. Also, I realized that my initial code description might not be exactly what you're looking for. So, if you could take a look and add your description, that'd be awesome! Thanks!

config/config_tmpl.go
xenbyte: Looks better.
I've commited the necessary changes

PR #493, opened 2023-09-02
[validators.http] Add last modified header validation

### LINKED ISSUE #414: Support for checking and comparing value in Last-Modified Header
**Support for checking and comparing Last-Modified Header**

I would like to see an additional validation based on Last Modified Header where I compare the value against the input value. It should return a successful validation if the difference is within the specified input duration else a failed validation. Based on documentation, success_header option in validator checks for specific time stamp and not the difference between current time and last modified time

robpickerill:
This PR adds support for validating if the last modified header is within the specified time duration.

Resolves #414


validators/http/http.go
```diff
@@ -230,5 +236,18 @@ func (v *Validator) Validate(input interface{}, unused []byte) (bool, error) {
 		}
 	}
 
+	if v.lastModifiedDiff != time.Duration(0) {
+		lastModified, err := time.Parse(time.RFC1123, res.Header.Get("Last-Modified"))
+		if err != nil {
+			v.l.Warningf("HTTP validation failure: Error parsing Last-Modified header: %v", err)
```
robpickerill, line 242: I fail the validation here on the parse error, on the presumption that anyone using this validation expects a valid header value.

syd20: I would agree that in this use case the Last Modified header needs a value else I would rather only check for the existence of the header using the other already available configuration options.

validators/http/proto/config.proto
```diff
@@ -39,4 +39,11 @@ message Validator {
   // Failure Header:
   //   If HTTP response headers match failure_header, validation fails.
   optional Header failure_header = 4;
+
+  // Last Modified Difference:
+  //   If specified, HTTP response's Last-Modified header is checked to be
+  //   within the specified time difference from the current time. Example:
+  //     last_modified_diff_seconds: 3600
+  //   This will check that the Last-Modified header is within the last hour.
+  uint64 last_modified_diff_seconds = 5;
```
manugarg, line 48: I am wondering if we should call it `max_last_modified_diff_sec`. Adding `max` makes it more explicit, and `sec` is what we seem to use everywhere else.

robpickerill: that makes sense, renamed to be consistent

manugarg (approved):

validators/http/http.go
```diff
@@ -37,6 +38,7 @@ type Validator struct {
 	failureStatusCodeRanges []*numRange
 	successHeaderRegexp     *regexp.Regexp
 	failureHeaderRegexp     *regexp.Regexp
+	maxLastModifiedDiffSec  time.Duration
```
manugarg, line 41: Since it's already parsed time duration, I'd remove Sec from the field name.
```suggestion
	maxLastModifiedDiff  time.Duration
```

manugarg (commented):
One minor thing, I missed before approving.

robpickerill:
My apologies Manu, I had missed the suggestion – now accepted and merged master, I think you should be good to go now, thanks!

robpickerill:
Patched, and you're now good to go! 👍

manugarg (approved):

manugarg:
> My apologies Manu, I had missed the suggestion – now accepted and merged master, I think you should be good to go now, thanks!
> 
> Edit: let me quickly patch up those tests.

Thanks! No worries.

PR #525, opened 2023-09-13
add changes to override GCP resources api endpoint

shellyvilenko:
Add ability to override gcp resources api endpoint.


rds/gcp/proto/config.proto
```diff
@@ -33,6 +33,9 @@ message GCEInstances {
   // only to the zones in the "us-east1" region.
   optional string zone_filter = 1;
 
+  // Only for testing.
+  optional string api_endpoint = 2;
```
manugarg, line 37: Can you please move this field to the ProviderConfig message:
https://github.com/cloudprober/cloudprober/blob/1e3decc83a66161fce1ad274c4a88b4e2bb089e9/rds/gcp/proto/config.proto#L83

This field will apply to all GCP resources so it makes sense for it exist at the GCP provider level. Note: If you don't want to modify other GCP resources to make use of this new field, you can just say in the comments (in proto) that this field is supported only for GCE instances for now.

It will be similar to the api_version field, so it will make sense to define it right next to it.

Once you define it there, you'll have to modify gcp.go in the same directory and pass this information to newGCEInstances().

manugarg (commented):
Please see my inline comment below.

You'll also need to run `tools/gen_pb_go.sh` for any changes to the protobuf files.

rds/gcp/proto/config.proto
shellyvilenko: Thanks for the prompt reply!
I fixed your comments.

manugarg (approved):

manugarg:
@shellyvilenko thanks for the PR!

PR #530, opened 2023-09-15
[alerting] Add alerts dashboard.

manugarg:


manugarg:
This is how alerts dashboard will look like:

image

PR #536, opened 2023-09-19
[config] Provide a way to hide secret environment variables.

manugarg:
- Provide an option to specify secret environment variables, using envSecret keyword.
- When such variables are present, /config-running will be automatically disabled.
- Add another page /config-parsed that shows the parsed config and makes it clear which variables are secret.


manugarg:
Screenshots:

image<br>
image<br>
image

Issue #535, opened 2023-09-19
HTTP probes do not recognize change of IP


ls692:
**Describe the bug**
If a probe target VM is recreated quickly enough and ends up with a new probe IP, HTTP probe still uses the old IP. 

**Cloudprober Version**
All versions of cloudprober are affected.

**Context**
HTTP probes use endpoint.Key() to determine targets that are active and ones need to be recreated
https://github.com/cloudprober/cloudprober/blob/master/probes/http/http.go#L524

endpoint,Key() is defined as Name + labels, excluding IP. https://github.com/cloudprober/cloudprober/blob/master/targets/endpoint/endpoint.go#L41

So if only VM IP changes, we end up reusing the same target but keep probing the old IP.

One simple fix is to modify endpoint.Key() to use the endpoint IP too. I will do some digging to see if that causes other complications. If so, another quickfix is to modify http.go to append IP to endpoint.Key().

manugarg:
Hi Lenin :wave: 😄 ,

Great find. I think this bug must have existed since [Oct, 2022](https://github.com/cloudprober/cloudprober/commit/03cda470703acec464edfa9171b7bc449298fb93), when we started switching to the model of storing IPs in the endpoint struct itself, instead of asking the "targets" to resolve it.

I think modifying endpoint.Key() to include IP should be okay.

manugarg:
Thanks @ls692 for fixing this in #538.

PR #538, opened 2023-09-20
[bugfix] Add IP to endpoint key. (#535)

ls692:
HTTP probes use endpoint key as identifier to decide when to recreate probes. During VM recreations, it is possible for VM name to stay the same but VM IP changes. By adding IP to key, we ensure that in such cases, HTTP probes are recreated.

### DIFF
--- a/targets/endpoint/endpoint_test.go
+++ b/targets/endpoint/endpoint_test.go
@@ -72,6 +72,7 @@ func TestKey(t *testing.T) {
 		ep := Endpoint{
 			Name:   test.name,
 			Port:   test.port,
+			IP:     test.ip,
 			Labels: test.labels,
 		}
 		t.Run(fmt.Sprintf("%v", ep), func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


targets/endpoint/endpoint_test.go
```diff
@@ -45,24 +46,27 @@ func TestKey(t *testing.T) {
...
 		},
 		{
 			name:   "t1",
 			port:   80,
 			labels: map[string]string{"dc": "xx", "app": "cloudprober"},
-			key:    "t1_80_app:cloudprober_dc:xx",
+			key:    "t1__80_app:cloudprober_dc:xx",
 		},
 	} {
 		ep := Endpoint{
```
manugarg, line 72: I think the problem is here. Endpoint is not using the test.ip field.

ls692:
Thanks manu, i missed copying that line of code from the internal repo. It is fixed now.

targets/endpoint/endpoint_test.go
ls692: Fixed now.

PR #543, opened 2023-09-30
[build] Test example configs during build

robpickerill:
This PR uses `-configtest` in CI to check that the example configs are valid.


tools/test_config.sh
manugarg, line 1: Can you please rename this file to test_example_configs?

manugarg (commented):
This is cool. I am surprised that most configs actually work. Thanks @robpickerill :)

One minor suggestion regarding the file name.

tools/test_config.sh
robpickerill: no problems, done, thanks Manu

Issue #293, opened 2023-02-27
Add a test to verify that all probe types allow empty configs.


manugarg:
TCP probe crashes if we don't provide a config (that's because we try to access a field of the config)  -- fix #292.  We ran into this problem with HTTP probe sometime back. We should a add a test that verifies that all probe types allow empty configs.

robpickerill:
Manu, would you be able to check that #544 achieves what you wanted here. Thanks!

manugarg:
@robpickerill Thanks for implementing this.

PR #544, opened 2023-09-30
[probes.test] Add tests to verify that empty configs work.

robpickerill:
This PR adds a test for empty configs, as defined in #293.


robpickerill:
CI fails due to:

> go: remove C:\Users\RUNNER~1\AppData\Local\Temp\go-build3371302423\b724\external.test.exe: Access is denied.

I'll look into this later

manugarg:
Thanks @robpickerill! This strategy looks good to me.

> CI fails due to:
> 
> go: remove C:\Users\RUNNER~1\AppData\Local\Temp\go-build3371302423\b724\external.test.exe: Access is denied.
> 
I looked at this on a Windows machine. I think I've an idea what the problem is. #547 should fix it.

probes/probes_unit_test.go
manugarg, line 1: Any reason to keep these tests in a dedicated file, instead of just in probes_test.go?

robpickerill: For this test, I put it inside the probes package as I'd used the internal function `initProbe` from the probes package here: https://github.com/cloudprober/cloudprober/blob/b271acc59315470a746076a4381f93ec3e21166f/probes/probes_unit_test.go#L31.

The other test files package name was probes_test, so its only got access to the exported functions.

Would you want to change this, we could merge into a single test file under the probes package.

manugarg: Ah, right. Sorry, missed that. Yeah, I remember keeping probes and targets test in different packages to be able to test extensions. Let's leave that for now.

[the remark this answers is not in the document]
robpickerill: Thanks Manu, my bad I forgot about the copyright headed. Let me add, and fix up this file. Will have it done in a few hours, and get it over to you.

manugarg (approved):

PR #587, opened 2023-10-20
[alerting] Move alerting outside of probes

manugarg:
- It's codebase and application is quite independent.
- Hide alerting under internal to avoid creating accidental dependency on its code.


manugarg:
I am not super-sure of putting this code under internal, but it has to be outside of probes.

Issue #576, opened 2023-10-18
ping probe doesn't increment "total" metric if the target name can't be resolved


jinmeiib:
**Describe the bug**

If the target of a ping probe is a domain name that can't be resolved, the probe reports the "total" metric that is not increasing. It's different from, e.g., tcp or http probes. It's not clear to me if this difference is intentional, but it would be better if the behavior in this case is consistent among different types of probes.

**Cloudprober Version**

0.13.0

**To Reproduce**

Run cloudprober with the following configuration. The target name `nosuchhost.jinmei.org` doesn't exist and can't be resolved.

```
probe {
  name: "HTTP probe"
  type: HTTP
  targets {
     host_names: "nosuchhost.jinmei.org"
  }
  http_probe {
    headers: [{
      name: "Host"
      value: "www.jinmei.org"
    }]
    method: GET
    protocol: HTTPS
  }
  interval_msec: 5000  # 5s
  timeout_msec: 1000   # 1s
  stats_export_interval_msec: 5000
}
probe {
  name: "TCP probe"
  type: TCP
  targets {
     host_names: "nosuchhost.jinmei.org"
  }
  tcp_probe {
    port: 80
  }
  interval_msec: 5000  # 5s
  timeout_msec: 1000   # 1s
  stats_export_interval_msec: 5000
}
probe {
  name: "PING probe"
  type: PING
  targets {
     host_names: "nosuchhost.jinmei.org"
  }
  interval_msec: 5000  # 5s
  timeout_msec: 1000   # 1s
  stats_export_interval_msec: 5000
}
```
Then the following logs will be emitted:
```
time=2023-10-18T14:20:48.346-07:00 level=WARN source=probes/tcp/tcp.go:178 msg="Target:nosuchhost.jinmei.org, doTCP: dial tcp: lookup nosuchhost.jinmei.org: no such host" system=cloudprober probe="TCP probe"
cloudprober 1697664045006931001 1697664048 labels=ptype=tcp,probe=TCP probe,dst=nosuchhost.jinmei.org total=1 success=0 latency=0.000
...
time=2023-10-18T14:20:46.678-07:00 level=WARN source=probes/http/http.go:290 msg="Get \"https://nosuchhost.jinmei.org\": dial tcp: lookup nosuchhost.jinmei.org: no such host" system=cloudprober probe="HTTP probe" target=nosuchhost.jinmei.org url=https://nosuchhost.jinmei.org
[...]

PR #620, opened 2023-11-01
[probes.ping] Count resolve/write failures as probe failures

manugarg:


manugarg:
This should address #576.

PR #632, opened 2023-11-08
Fix Dockerfile.dev

jumpojoy:
* Update golang
* Update location of cloudprober_test.cfg


Dockerfile.dev
```diff
@@ -1,13 +1,13 @@
 # Dockerfile for quick development and testing
 # To build:
 #    docker build -t cloudprober:test . -f Dockerfile.dev
-FROM golang:1.18-alpine as build
+FROM golang:1.21.1-alpine as build
```
manugarg, line 5:
```suggestion
FROM golang:1.21-alpine as build

```

PR #634, opened 2023-11-09
Use network buffer less aggressive in ping probe

jumpojoy:
When have higher number of targets (seen on 600+), no matter in single
probe or in different probes, we may cause network buffer overflow as
we add packets to the buffer much faster than NIC can handle them.
To avoid this we may increase net.core.wmem sysctl, or add 1 milisecond
dalay between buffer writes (which is done by this patch). This allows
to have stable results for even 2k targets in single probe.
    
Also patch improves logging
   - Added debug log when probe is started.
   - Do not print duplicate p.name as it is always added by logger,
      add p.runCnt instead
   - Improve debug message for a case when packet does not match runId
      this is usually happening when we received packet after probe timeout.

### DIFF
--- a/probes/ping/ping.go
+++ b/probes/ping/ping.go
@@ -301,8 +301,8 @@ func (p *Probe) sendPackets(runID uint16, tracker chan bool) {
 			}
 
 			tracker <- true
-			// When have high number targets (does not matter in one probe or in different probes),
-			// do not send packets to aggressive to avoid buffer overflow.
+			// Sleep between pushing packets to avoid network buffer overflow
+			// in case of larger number of targets.
 			time.Sleep(1 * time.Millisecond)
 		}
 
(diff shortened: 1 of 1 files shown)


manugarg:
Thanks @jumpojoy! 

This changes makes sense to me. Can you please sync your repo and send it again? There is a Dockerfile.dev in the PR causing conflicts.

jumpojoy:
Thanks for review @manugarg, done.

probes/ping/ping.go
```diff
@@ -301,6 +301,9 @@ func (p *Probe) sendPackets(runID uint16, tracker chan bool) {
 			}
 
 			tracker <- true
+			// When have high number targets (does not matter in one probe or in different probes),
+			// do not send packets to aggressive to avoid buffer overflow.
```
manugarg, line 305:
```suggestion
			// Sleep between pushing packets to avoid network buffer overflow
			// in case of larger number of targets.
```

probes/ping/ping.go
```diff
@@ -426,7 +430,7 @@ func (p *Probe) recvPackets(runID uint16, tracker chan bool) {
 
 		// check if this packet belongs to this run
 		if !matchPacket(runID, pkt.id, pkt.seq, p.useDatagramSocket) {
-			p.l.Info("Reply ", pkt.String(rtt), " Unmatched packet, probably from the last probe run.")
+			p.l.Info("Reply ", pkt.String(rtt), " Unmatched packet, probably received after probe timeout is reached.")
```
manugarg, line 433:
```suggestion
			p.l.Info("Reply ", pkt.String(rtt), " Unmatched packet, probably received after last probe's timeout.")
```

manugarg (commented):
Couple of minor suggestions.

PR #654, opened 2023-12-19
[surfacers.otel] Fix a bug in exporter configuration logic.

manugarg:


TaylorMutch (approved):
Lgtm, I also ran into this issue when trying this out and the config was ignored. This should fix it

PR #666, opened 2024-01-02
gitignore: unignore checked-in files without extensions

AdamEAnderson:
Currently, gitignore is ignoring some root files that don't have extensions. This causes weird behavior since there are ignored files currently in the repo state.


manugarg:
Thanks!

Issue #645, opened 2023-12-14
Is there a way to write cloudprober probe config to persistent memory?


hgsat123:
We are using gRPC client to perform dynamic update (automated) of probe config so that cloudprober can continue the updated probe on the fly. We are not using configmap as probe configs are updated whenever required. Now, as i understand (correct me), cloudprober copies the (mounted configmap - probe to in-memory or allows gRPC client to write/update probe config directly on cloudprober in-memory cache so that it can read / access the probe on the fly. THis is perfectly working. However, if cloudprober restarts due to any resource issue or underlying host issue, then all the probe updated in its memory are reset? so probe will not happen. Is there a way other then mounting probe via configmap which auto mounts the same (fixed list of probe) every time whenever container restarts, to dynamically update cloudprober probe to persistent memory (like memcached or Radis) which cloudprobe can use to read its probe file?  Just thought of asking this question.

manugarg:
@hgsat123 Currently, there is no such way. The idea was whichever system is programming Cloudprober, will also take care of restarted processes and programming them again.

Saving the current config to the local disk (mounted volume) on every gRPC programming will be easier than implementing state in management memcached or redis. I'll give the former more thought.

manugarg:
I'm trying something here: https://github.com/cloudprober/cloudprober/pull/671/files

If you provide a flag: --config_save_path, Cloudprober will automatically save config at that path after every change. You can also save config at an arbitrary path through an RPC call (currently named SaveConfigToDisk).

hgsat123:
ok. Thanks manu. So, is this --config_save_path < mounted-file-path> helps to save back whatever config there in In memory cache back on to this saved file? This would be very helpful as we don't use configmap to mount a fixed probe config, whatever probe config written dynamically to CloudProber in memory would be synced with file which can be reloaded to In memory incase of container restarts??  Let me know whenever this fix / experimental flag checked -in. Will test.

manugarg:
Sorry for taking so long to merge #671. 

From my commit message:
```
If you start cloudprober with flag --probes_config_save_path, it will automatically save new probes config to the provided path on any dynamic programming through gRPC.

You can in turn include this saved config in the main config file ([include_example](https://github.com/cloudprober/cloudprober/tree/master/examples/include), Cloudprober will reload the whole config on restart.
```

One way to do this:
```
cloudprober --config_file=$HOME/cloudprober.cfg --probes_config_save_path=save_config.cfg

In cloudprober.cfg:

[...]

PR #671, opened 2024-01-11
[config] Provide a mechanism to save config to disk

manugarg:
If you start cloudprober with flag `--probes_config_save_path`, it will automatically save new probes config to the provided path on any dynamic programming through gRPC.


manugarg:
Fixes https://github.com/cloudprober/cloudprober/issues/645.

PR #673, opened 2024-02-01
[surfacers.stackdriver] Rollback the disable overwrite option.

manugarg:
= We found out in https://github.com/cloudprober/cloudprober/issues/667#issuecomment-1920571733 that stackdriver doesn't allow multiple datapoints for a timeseries in a single call, no matter how we send it.
= Disabling data overwrite in cache will result in unnecessary confusion and noisy logs. Hence, rolling back this option.


v-pratap:
Please do not revert this PR, we will use this feature for benchmarks prober (high throughput). We are mostly planning to precalculate the metrics and then emit the metrics but still there will be overwrite issue.

manugarg:
> Please do not revert this PR, we will use this feature for benchmarks prober (high throughput). We are mostly planning to precalculate the metrics and then emit the metrics but still there will be overwrite issue.

@v-pratap Not overwriting is never going to work. Not overwriting data means keeping more than 1 datapoint for timeseries in cache. 

```
cache[timeseries] -> point
vs
cache[timeseries] -> []point
```
Problem if even if we keep more than one datapoint per timeseries in cache, stackdriver is not going to accept it. 

Does that make sense?

Also, it's not reverting the whole of the last change. Previous change fixed a bug in the map values handling, which must have also been impacting you. That fix will remain.

v-pratap:
Let's suppose the batch_timer_sec is more than 10second (eg: 20 second no duplicate metrics in first 10 second and no duplicate in 10 to 20 seconds but there are duplicates in 0 to 20 seconds )  , then in that case stackdriver should not overrite the metrics?

manugarg:
> Let's suppose the batch_timer_sec is more than 10second (eg: 20 second no duplicate metrics in first 10 second and no duplicate in 10 to 20 seconds but there are duplicates in 0 to 20 seconds ) , then in that case stackdriver should not overrite the metrics?

Stackdriver is rejecting timeseries because of multiple data points per call. So even if batch timer is larger than 10s, the call will be rejected if there are multiple data points per timeseries. See: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries
image

I was thinking sending multiple data points in different timeseries objects will be okay, but it seems that stackdriver normalized the data before accepting it.

v-pratap:
Hi @manugarg, I understand this. Can you please wait for few days before merging this rollback?

manugarg:
> Hi @manugarg, I understand this. Can you please wait for few days before merging this rollback?

Hi @v-pratap, I am going to submit it now. If you find counter evidence, we can rollback this PR.

PR #697, opened 2024-03-14
Fix client cert handling when cert reloading is enabled

cbroglie:
Since cloudprober is the client, the dynamic cert loading needs to be implemented in GetClientCertificate rather than GetCertificate.


manugarg (approved):
Thanks for fixing this, Chris!

cbroglie:
No problem, sorry I didn't around to testing til now!

markoposavec:
This change breaks how certificate reloading works for grpc server, since in that context cloudprober is a server and it uses the same UpdateTLSConfig function that you changed 😢

manugarg:
https://github.com/cloudprober/cloudprober/pull/719 should add reloading behavior to server side certificates as well.

PR #707, opened 2024-03-29
Provide DNSOverride for targets

sfc-gh-raram:
The default behavior of cloudprober is to use the local DNS resolver for all targets. While this covers most use cases, some times we want to resolve a target using a DNS resolver that we provide such as Cloudflare's DNS resolver (1.1.1.1). This PR allows the user to specify the dns resolver for a probe and for static targets, it will use the specified dns resolver.

This change includes a unit test and was tested in a k8s environment where the cloudprober config included probes that used the dns resolver override and the default dns resolver.


manugarg (commented):
Hey Rahul,

Thanks for sending this. I think being able to override DNS resolver makes a lot of sense. I was wondering if you've considered adding the new option to TargetsDef proto instead:
https://github.com/cloudprober/cloudprober/blob/24587fb522f6beeddd9e48f2f6ef4078fbd41ac3/targets/proto/targets.proto#L100

I think that's a better place for it and it will simplify the implementation quite a bit as we'll not have to pass this option around.

sfc-gh-raram:
> Hey Rahul,
> 
> Thanks for sending this. I think being able to override DNS resolver makes a lot of sense. I was wondering if you've considered adding the new option to TargetsDef proto instead:
> 
> https://github.com/cloudprober/cloudprober/blob/24587fb522f6beeddd9e48f2f6ef4078fbd41ac3/targets/proto/targets.proto#L100
> 
> I think that's a better place for it and it will simplify the implementation quite a bit as we'll not have to pass this option around.

Agreed. Updated the PR with the new changes. Tested them locally and in k8s. When you get a chance, could you take another look? Thank you.

targets/resolver/resolver.go
```diff
@@ -198,6 +203,29 @@ func NewWithResolve(resolveFunc func(string) ([]net.IP, error)) *Resolver {
...
 
+func resolveFuncDNSOverride(host string) ([]net.IP, error) {
+	r := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			d := net.Dialer{
+				Timeout: defaultMaxAge,
+			}
+			publicDNSIP := dnsOverride + ":53"
+			return d.DialContext(ctx, network, publicDNSIP)
```
manugarg, line 214: Can you call it dnsIP as it may or may not be public DNS.

targets/resolver/resolver.go
```diff
@@ -198,6 +203,29 @@ func NewWithResolve(resolveFunc func(string) ([]net.IP, error)) *Resolver {
...
+}
+
+func NewOverrideResolver(dnsResolverOverride string) *Resolver {
+	resolveFunc := net.LookupIP
+	if dnsResolverOverride != "" {
+		dnsOverride = dnsResolverOverride
+		resolveFunc = resolveFuncDNSOverride
+	}
+	return NewWithResolve(resolveFunc)
+}
```
manugarg, line 227: I think storing dns override in a global variable is not safe. Why not just do something like this:

```
func NewWithOverrideResolver(dnsResolverOverride string) *Resolver {
        resolveFunc = func(host string) ([]net.IP, error) {
       	   r := &net.Resolver{
		PreferGo: true,
		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
			d := net.Dialer{
				Timeout: defaultMaxAge,
			}
			dnsIP := dnsOverride + ":53"
			return d.DialContext(ctx, network, dnsIP)
		},
	   }
	   return r.LookupIP(context.Background(), "ip", host)
        }
	return NewWithResolve(resolveFunc)
}
```

targets/targets.go
```diff
@@ -48,8 +48,11 @@ import (
 // globalResolver is a singleton DNS resolver that is used as the default
 // resolver by targets. It is a singleton because dnsRes.Resolver provides a
 // cache layer that is best shared by all probes.
+
+// DNSResolverOverrides is a cache of the dns resolver overrides
 var (
-	globalResolver *dnsRes.Resolver
+	globalResolver       *dnsRes.Resolver
+	dnsResolverOverrides = make(map[string]*dnsRes.Resolver)
```
manugarg, line 55: I think we can skip sharing override DNS resolvers. They will be less enough that we won't benefit much from sharing, also to maintain this global map we'll have to use mutexes.

targets/targets.go
```diff
@@ -430,6 +436,22 @@ func SetSharedTargets(name string, tgts Targets) {
...
+	dnsResolverOverride := targetsDef.GetDnsResolverOverride()
+	if dnsResolverOverride == "" {
+		return globalResolver
+	}
+
+	_, isDNSResolverPresent := dnsResolverOverrides[dnsResolverOverride]
+	if !isDNSResolverPresent {
+		dnsResolverOverrides[dnsResolverOverride] = dnsRes.NewOverrideResolver(dnsResolverOverride)
+	}
+	return dnsResolverOverrides[dnsResolverOverride]
```
manugarg, line 452: See my comment about sharing override resolvers. I think we can just create a new override resolver and use it:

```
return dnsRes.NewOverrideResolver(dnsResolverOverride)
```

In fact, this block can simply be part of the New function.

targets/targets.go
```diff
@@ -323,9 +327,11 @@ func New(targetsDef *targetspb.TargetsDef, ldLister endpoint.Lister, globalOpts
 		return nil, fmt.Errorf("targets.New(): Error making baseTargets: %v", err)
 	}
 
+	targetDNSResolver := createOverrideDNSResolver(targetsDef)
+
```
manugarg, line 331: Just create this resolver before calling baseTargets and pass it on to baseTargets?

manugarg (commented):
Hi Rahul,

Few more comments. I am out of office until end of the week, so take your time.

Thanks,
Manu

targets/targets.go
sfc-gh-raram: There were a couple of things I was thinking about when handling this issue. First, it's posisble to overwrite the resolver after basetargets is called such as in [static_targets.go](https://github.com/cloudprober/cloudprober/blob/master/targets/statictargets.go#L81). To fix this, I modified the static target method signature to include the resolver. I also substitued the other methods that used to take the globalResolver with the targetResolver variable.

I'm also already calling the createOverrideDNSResolver method in base targets instead of setting it to the global resolver because some targets types do not set the resolver. Is this right way to approach this?

sfc-gh-raram:
I also have a question regarding the unit test. The unit test checks whether the correct DNSIP is present in the [targets struct](https://github.com/cloudprober/cloudprober/blob/c3943aa8013324a9b4f87f27b8d643bec984ab9f/targets/targets.go#L128). Ideally, I would like to put this field in the [resolver struct](https://github.com/cloudprober/cloudprober/blob/master/targets/resolver/resolver.go#L42). However, I run into the issue that the resolver struct does not implement the [Resolver interface](https://github.com/cloudprober/cloudprober/blob/master/targets/endpoint/endpoint.go#L68) since the resolve method signatures are different, preventing me from casting targets.resolver into the resolver struct. Is there a way to fix this, or am I missing something? Thank you.

targets/targets.go
```diff
@@ -322,10 +324,11 @@ func New(targetsDef *targetspb.TargetsDef, ldLister endpoint.Lister, globalOpts
...
 		return nil, fmt.Errorf("targets.New(): Error making baseTargets: %v", err)
 	}
-
+	targetDNSResolver, dnsIP := createOverrideDNSResolver(targetsDef)
+	t.DnsIP = dnsIP
 	switch targetsDef.Type.(type) {
 	case *targetspb.TargetsDef_HostNames:
-		st, err := staticTargets(targetsDef.GetHostNames())
+		st, err := staticTargets(targetsDef.GetHostNames(), targetDNSResolver)
 		if err != nil {
```
manugarg, line 332: I was thinking instead of recreating targetDNSResolver, can we just use t.resolver in this call and the other calls below?

```suggestion
		st, err := staticTargets(targetsDef.GetHostNames(), t.resolver)
		if err != nil {
```

targets/targets.go
```diff
@@ -430,6 +433,14 @@ func SetSharedTargets(name string, tgts Targets) {
...
 }
 
+func createOverrideDNSResolver(targetsDef *targetspb.TargetsDef) (*dnsRes.Resolver, string) {
+	dnsResolverOverride := targetsDef.GetDnsResolverOverride()
+	if dnsResolverOverride == "" {
+		return globalResolver, ""
+	}
+	return dnsRes.NewWithOverrideResolver(dnsResolverOverride), dnsResolverOverride
+}
+
```
manugarg, line 443: If we call this function only once (see my comment above), I'd prefer that we inline this in baseTargets(...):
```
resolver := globalResolver
if ip := targetsDef.GetDns... ; ip != "" {
  resolver = dnsRes.New... 
}
```

targets/targets.go
manugarg: I think what you've done is right. staticTargets should take a resolver too, same for other targets types.

manugarg:
> I also have a question regarding the unit test. The unit test checks whether the correct DNSIP is present in the [targets struct](https://github.com/cloudprober/cloudprober/blob/c3943aa8013324a9b4f87f27b8d643bec984ab9f/targets/targets.go#L128). Ideally, I would like to put this field in the [resolver struct](https://github.com/cloudprober/cloudprober/blob/master/targets/resolver/resolver.go#L42). However, I run into the issue that the resolver struct does not implement the [Resolver interface](https://github.com/cloudprober/cloudprober/blob/master/targets/endpoint/endpoint.go#L68) since the resolve method signatures are different, preventing me from casting targets.resolver into the resolver struct. Is there a way to fix this, or am I missing something? Thank you.

Ha, that's surprising. You're right.. adding a DNS IP field to resolver makes more sense, then you should be able to access it from target's resolver by type-casting.

Does the following not work? 
```
in resolver.go:
struct Resolver {
   ...
   DNSOverrideIP string // Used for testing
}

in targets_test.go:

...
r, ok := t.resolver.(*dnsRes.Resolver)
if !ok {
  t.Fatalf(..)
}
assert.Equal(t, wantDNSOverrideIP, r.DNSOverrideIP, ..)
...
```

targets/proto/targets.proto
```diff
@@ -187,6 +187,9 @@ message TargetsDef {
   // are specified.
   optional bool exclude_lameducks = 22 [default = true];
 
+  // Provide a dns resolver override instead of using the default dns resolver.
+  optional string dns_resolver_override = 37;
+
```
manugarg, line 192: What do you think about calling it just dns_server:

```suggestion
  // Use this dns server instead of the default local DNS server
  optional string dns_server = 37;

```

targets/targets.go
sfc-gh-raram: I think we would need to cast since t.resolver is of type endpoint.Resolver while the methods below want an object of type *dnsRes.Resolver.

targets/targets.go
sfc-gh-raram: This makes sense to me. I can inline this in the new method and pass it into baseTargets to avoid unnecessary casting (see my comment above).

sfc-gh-raram:
> > I also have a question regarding the unit test. The unit test checks whether the correct DNSIP is present in the [targets struct](https://github.com/cloudprober/cloudprober/blob/c3943aa8013324a9b4f87f27b8d643bec984ab9f/targets/targets.go#L128). Ideally, I would like to put this field in the [resolver struct](https://github.com/cloudprober/cloudprober/blob/master/targets/resolver/resolver.go#L42). However, I run into the issue that the resolver struct does not implement the [Resolver interface](https://github.com/cloudprober/cloudprober/blob/master/targets/endpoint/endpoint.go#L68) since the resolve method signatures are different, preventing me from casting targets.resolver into the resolver struct. Is there a way to fix this, or am I missing something? Thank you.
> 
> Ha, that's surprising. You're right.. adding a DNS IP field to resolver makes more sense, then you should be able to access it from target's resolver by type-casting.
> 
> Does the following not work?
> 
> ```
> in resolver.go:
> struct Resolver {
>    ...
>    DNSOverrideIP string // Used for testing
> }
> 
> in targets_test.go:
> 
> ...
> r, ok := t.resolver.(*dnsRes.Resolver)
> if !ok {
>   t.Fatalf(..)
> }
> assert.Equal(t, wantDNSOverrideIP, r.DNSOverrideIP, ..)
> ...
> ```

Unfortunately, it does not work. During runtime, I receive the following error `panic: interface conversion: endpoint.Resolver is *targets.targets, not *resolver.Resolver [recovered]`. I suspect it's because the [resolve function in the resolver struct](https://github.com/cloudprober/cloudprober/blob/master/targets/resolver/resolver.go#L46) does not include the ipVert int argument included in the [Resolve function of the Resolver interface](https://github.com/cloudprober/cloudprober/blob/master/targets/endpoint/endpoint.go#L68).

targets/targets.go
```diff
@@ -317,15 +319,19 @@ func rdsClientConf(pb *targetspb.RDSTargets, globalOpts *targetspb.GlobalTargets
...
+	resolver := globalResolver
+	ip := ""
+	if ip = targetsDef.GetDnsServer(); ip != "" {
+		resolver = dnsRes.NewWithOverrideResolver(ip)
+	}
+	t, err := baseTargets(targetsDef, ldLister, l, resolver, ip)
 	if err != nil {
 		globalLogger.Error("Unable to produce the base target lister")
 		return nil, fmt.Errorf("targets.New(): Error making baseTargets: %v", err)
 	}
```
manugarg, line 331: One (hopefully) minor suggestion:
Instead of passing resolver and ip to baseTargets, can you please set it after baseTargets returns:

(that way you can leave base targets signature unchanged)

```
t, err := baseTargets(....)
if err ..

t.resolver = resolver
```

(you'll have to do the same in StaticTargets() function too -- or, maybe you let baseTargets set resolver to globalResolver by default, that StaticTargets can be simple)

targets/targets.go
```diff
@@ -317,15 +319,19 @@ func rdsClientConf(pb *targetspb.RDSTargets, globalOpts *targetspb.GlobalTargets
...
-
 	switch targetsDef.Type.(type) {
 	case *targetspb.TargetsDef_HostNames:
-		st, err := staticTargets(targetsDef.GetHostNames())
+		st, err := staticTargets(targetsDef.GetHostNames(), resolver, ip)
 		if err != nil {
```
manugarg, line 335: Following the suggestion above, you can skip passing resolver and ip to staticTargets, and then assign `st` to just the `t.lister`, i.e. leave t.resolver unchanged.

manugarg (commented):
Looks good. Just last couple of minor comments if they make sense to you. Otherwise we can go with this too.

manugarg (approved):
LGTM. Thanks!

PR #711, opened 2024-04-05
Postgres surfacer metric filter bugfix

markoposavec:
Hello,
during the initialisation of postgres surfacer options are never set but are later on used to check if metrics should be sent to Postgres database.
I added the options as a parameter of creating postgres surfacer.


manugarg:
Thanks so much @markoposavec for the fix!

PR #717, opened 2024-04-10
[surfacer.postgres] Postgresql surfacer bulk inserts

markoposavec:
Hello!
I've added option to add bulk writing into postgres database. Currently all metrics from one metrics event object are written in one transaction which was very slow in my use case because of slow connections towards database. I've added batching of metrics events into one transaction. This change improved the speed in my use case by ~10x.

With default values system is working as it was before, event metrics are sent one by one to database.

Let me know if I should change anything,
Thanks!


surfacers/internal/postgres/proto/config.proto
```diff
@@ -26,6 +26,15 @@ message SurfacerConf {
...
   optional int64 metrics_buffer_size = 3 [default = 10000];
+
+  // Minimum number of event metric objects to flush in a single batch. Event metrics
+  // will be flushed when this number of event metric objects are available or when
+  // the flush interval is reached, whichever happens first. Default value is 1.
+  optional int64 metrics_batch_minimum_flush_size = 5;
+
+  // Maximum interval between flushes, in milliseconds. Every time batch is
+  // flushed, this timer is reset. Default value is 1000.
+  optional int64 metrics_batch_flush_interval_msec = 6;
```
manugarg, line 37: Can you please use the following variables instead. These are pretty much standard across all surfacers:
https://github.com/cloudprober/cloudprober/blob/master/surfacers/internal/datadog/proto/config.proto

```
  // The maximum number of metrics that will be published at one
  // time. Metrics will be stored locally in a cache until this 
  // limit is reached. Datadog's SubmitMetric API has a maximum payload 
  // size of 500 kilobytes (512000 bytes). Compressed payloads must have a 
  // decompressed size of less than 5 megabytes (5242880 bytes).
  // Metrics will be published when the timer expires, or the buffer is
  // full, whichever happens first.
  optional int32 metrics_batch_size = 5 [default = 1000]; 

  // The maximum amount of time to hold metrics in the buffer (above).
  // Metrics will be published when the timer expires, or the buffer is
  // full, whichever happens first. 
  optional int32 batch_timer_sec = 6 [default = 30];
```

surfacers/internal/postgres/postgres.go
```diff
@@ -275,18 +292,44 @@ func (s *Surfacer) init(ctx context.Context) error {
...
+
+}
+
+func resolveMetricsBatchFlushInterval(metricsBufferFlushIntervalMsec *int64, defaultBufferFlushIntervalMsec int64) int64 {
+	if metricsBufferFlushIntervalMsec == nil {
+		return defaultBufferFlushIntervalMsec
+	}
+	return *metricsBufferFlushIntervalMsec
+}
+
```
manugarg, line 332: These are very simple "if not / default" rules, I think it will be better to just inline them

surfacers/internal/postgres/postgres.go
```diff
@@ -275,18 +292,44 @@ func (s *Surfacer) init(ctx context.Context) error {
...
-				if err := s.writeMetrics(em); err != nil {
-					s.l.Warningf("Error while writing metrics: %v", err)
+				buffer = append(buffer, em)
+				if int64(len(buffer)) >= metricsBatchMinimumFlushSize {
+					if err := s.writeMetrics(buffer); err != nil {
+						s.l.Warningf("Error while writing metrics: %v", err)
+					}
+					buffer = buffer[:0]
+					wait = time.After(flushInterval)
+				}
```
manugarg, line 302:
```suggestion
				if int64(len(buffer)) >= metricsBatchMinimumFlushSize {
					if err := s.writeMetrics(buffer); err != nil {
						s.l.Warningf("Error while writing metrics: %v", err)
					}
					buffer = buffer[:0]
					flushTicker.Reset()
				}
```

surfacers/internal/postgres/postgres.go
```diff
@@ -266,6 +275,14 @@ func (s *Surfacer) init(ctx context.Context) error {
...
 		defer s.db.Close()
 
+		metricsBatchMinimumFlushSize := resolveMetricsBatchMinimumFlushSize(s.c.MetricsBatchMinimumFlushSize, DefaultMetricsBatchMinimumFlushSize)
+		metricsBatchFlushIntervalMsec := resolveMetricsBatchFlushInterval(s.c.MetricsBatchFlushIntervalMsec, DefaultMetricsBatchFlushIntervalMsec)
+
+		buffer := make([]*metrics.EventMetrics, 0, metricsBatchMinimumFlushSize)
+		flushInterval := time.Duration(metricsBatchFlushIntervalMsec) * time.Millisecond
+
+		var wait = time.After(flushInterval)
+
```
manugarg, line 285: Instead of using a time.After channel and creating a new channel everytime, I'd suggest use a ticker here:

```suggestion
		flushTicker := time.Ticker(flushInterval)
		defer flushTicker.Stop()


```

surfacers/internal/postgres/postgres.go
```diff
@@ -275,18 +292,44 @@ func (s *Surfacer) init(ctx context.Context) error {
...
+			case <-wait:
+				if len(buffer) > 0 {
+					if err := s.writeMetrics(buffer); err != nil {
+						s.l.Warningf("Error while writing metrics: %v", err)
+					}
+					buffer = buffer[:0]
 				}
+				wait = time.After(flushInterval)
 			}
 		}
```
manugarg, line 312:
```suggestion
			case <-flushTicker.C:
				if len(buffer) > 0 {
					if err := s.writeMetrics(buffer); err != nil {
						s.l.Warningf("Error while writing metrics: %v", err)
					}
					buffer = buffer[:0]
				}
			}
		}
```

manugarg (commented):
Thanks @markoposavec for this PR. Please take a look at the inline comments.

surfacers/internal/postgres/proto/config.proto
markoposavec: Changed the variables to the ones suggested.

surfacers/internal/postgres/postgres.go
markoposavec: Removed the functions.

markoposavec:
For some reason made a rebase and it shows changes from the other commit about tls.

manugarg:
> For some reason made a rebase and it shows changes from the other commit about tls.

That's weird. Can you try the following:

```
git checkout master
git pull master
git checkout <your branch name>
git rebase master
git push -f
```

I am not sure if above will work, might need to check internet :)

markoposavec:
that fixed the issue, thanks 😄

surfacers/internal/postgres/postgres.go
```diff
@@ -266,6 +275,22 @@ func (s *Surfacer) init(ctx context.Context) error {
...
 
+		metricsBatchSize := DefaultMetricsBatchSize
+		if s.c.MetricsBatchSize != nil {
+			metricsBatchSize = *s.c.MetricsBatchSize
+		}
+
+		batchTimerSec := DefaultBatchTimerSec
+		if s.c.BatchTimerSec != nil {
+			batchTimerSec = *s.c.BatchTimerSec
+		}
```
manugarg, line 286: Can you please kill the default variables and do this:

```
metricsBatchSize, batchTimerSec := s.c.GetMetricsBatchSize(), s.c.GetBatchTimerSec()
```

Protobufs make sure that default values work.

manugarg (commented):
One minor suggestion. Looks good otherwise.

surfacers/internal/postgres/postgres.go
markoposavec: wasn't aware that this exists 😄 
changed it

PR #744, opened 2024-05-14
Fix logical race condition in external server probe handling

apolcyn:
Currently there is a logical race condition as follows:

Suppose there is an external probe running in SERVER mode, with only one `target`.

1) cloudprober [sends the single request](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L268)

2) If the external probe server responds quickly enough, cloudprober can [read the reply](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L224), but then when it [checks](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L249) `sendDoneCh`, it might not be closed yet.

3) cloudprober's reader go-routine goes back to blocking on the [select](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L216) statement.

4) Now the parent go-routine [closes](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L273) `sendDoneCh`.

At this point, the reader go-routine will remain blocked on the top [select](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L216) statement.

It will remain there until the context times out.

-------

The impact of this bug is probably not catastrophic in most cases, but it causes a bunch of spurious "deadline exceeded" error logs from [here](https://github.com/cloudprober/cloudprober/blob/d4bfee74f61d59e79c043da3dd030ad6bcdc7f27/probes/external/external_server.go#L218) if nothing else.

### DIFF
--- a/probes/external/external_server.go
+++ b/probes/external/external_server.go
@@ -198,7 +198,6 @@ func (p *Probe) runServerProbe(ctx, startCtx context.Context) {
 
 	outstandingReqs := make(map[int32]requestInfo)
 	var outstandingReqsMu sync.RWMutex
-	sendDoneCh := make(chan struct{})
 
 	if err := p.startCmdIfNotRunning(startCtx); err != nil {
 		p.l.Error(err.Error())
@@ -212,6 +211,7 @@ func (p *Probe) runServerProbe(ctx, startCtx context.Context) {
 
 		// Read probe replies until we have no outstanding requests or context has
 		// run out.
+		expectedRepliesReceived := 0
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


manugarg:
Thanks @apolcyn!

Just to summarize for myself: race condition occurs when we are able to read the reply, but send channel is not closed so we come back to reading the reply channel which will not yield anymore as reply has already been read. That will cause goroutine to get blocked until timeout.

That makes sense to me. I'll take a further lok.

PR #781, opened 2024-06-18
Use context in ServeContext.

stvnrhodes:
I noticed that https://pkg.go.dev/github.com/cloudprober/cloudprober/probes/external/serverutils#ServeContext doesn't take a context. This has been true since it was added in https://github.com/cloudprober/cloudprober/pull/695, so it looks like an oversight. This change updates the function to properly take a context (and to work with its example code.


manugarg:
Ah, you're right. It does seem like an oversight. Thanks!

PR #783, opened 2024-06-22
Fixing getting the cert files form Google Storage

allagoichuk:
This cl https://github.com/cloudprober/cloudprober/pull/396 breaks accessing Google Storage as path.join converts double slash into single slash

Below code returns https:/storage.googleapis.com/bbmc-cloudradar-internal-ip-only/rds/ssl/cloudprober.crt
package main

```
import (
	"fmt"
	"path"
)

const gcsHTTPBaseURL = "https://storage.googleapis.com"

func main() {
	fmt.Printf(path.Join(gcsHTTPBaseURL, "bbmc-cloudradar-internal-ip-only/rds/ssl/cloudprober.crt"))
}
```

### DIFF
--- a/internal/file/gcs.go
+++ b/internal/file/gcs.go
@@ -33,7 +33,11 @@ func gcsRequest(ctx context.Context, method, objectPath string) (*http.Response,
 		return nil, err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, method, path.Join(gcsHTTPBaseURL, objectPath), nil)
+	u, _ := url.Parse(gcsHTTPBaseURL)
+	u.Path = path.Join(u.Path, objectPath)
+
+
+	req, err := http.NewRequestWithContext(ctx, method, u, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/file/gcs.go
+++ b/internal/file/gcs.go
@@ -33,7 +33,11 @@ func gcsRequest(ctx context.Context, method, objectPath string) (*http.Response,
 		return nil, err
 	}
 
-	u, _ := url.Parse(gcsHTTPBaseURL)
+	u, err := url.Parse(gcsHTTPBaseURL)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/file/gcs.go
```diff
@@ -33,7 +33,11 @@ func gcsRequest(ctx context.Context, method, objectPath string) (*http.Response,
 		return nil, err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, method, path.Join(gcsHTTPBaseURL, objectPath), nil)
+	u, _ := url.Parse(gcsHTTPBaseURL)
```
manugarg, line 36: I'd suggest panic'ing on URL parse error to catch any errors in gcsHTTPBaseURL over time.

```suggestion
	u, err := url.Parse(gcsHTTPBaseURL)
	if err != nil {
	  // This should never happen as base URL is defined by us.
	  panic("invalid GCS base URL: " + err.Error())
	}  
```

manugarg (commented):
@allagoichuk Many thanks for fixing this.

One minor suggestion but looks good otherwise.

manugarg:
Test failures are because of the missing `net/url` import.

internal/file/gcs.go
```diff
@@ -33,7 +34,15 @@ func gcsRequest(ctx context.Context, method, objectPath string) (*http.Response,
...
-	req, err := http.NewRequestWithContext(ctx, method, path.Join(gcsHTTPBaseURL, objectPath), nil)
+	u, err := url.Parse(gcsHTTPBaseURL)
+	if err != nil {
+	  // This should never happen as base URL is defined by us.
+	  panic("invalid GCS base URL: " + err.Error())
+	}  
+	u.Path = path.Join(u.Path, objectPath)
+
+
+	req, err := http.NewRequestWithContext(ctx, method, u, nil)
```
manugarg, line 45:
```suggestion

	req, err := http.NewRequestWithContext(ctx, method, u.String(), nil)
```

PR #793, opened 2024-07-04
Avoid emitting probe metrics after context cancellation

AdamEAnderson:
When a probe is stopped, i.e. via the probe removal process, sometimes the cancellation of the probe context happens while a probe run is in progress. This can cause that probe run to be interrupted mid-execution and return with a failure, which then gets reported through the metrics channel and out to the surfacer as a failed probe, even though the failure was due to probe cancellation rather than an external cause.

This PR adds a check to make sure the context is still live after the probe run is completed but before metrics are emitted. If the context is done, the probe stops without emitting the metrics.

Probe types and their susceptibilities for this issue:
- dns
  - dns does not have the issue because the probe run is not interrupted by context closure
- external (non-server)
  - this does have the issue if the context closes during the process execution as that will forcefully kill the process and result in a failure
- external (server)
  - this does not have the issue because the metrics reporting is bounded by the context, and because process failures are not emitted as event metrics
- grpc
  - this does not have the issue as metrics are emitted on the next loop after running probes, which involves checking the context again. Individual target cancellations are also handled because all target removals delete the probe result before the next metrics emission cycle runs.
- http
  - has this issue, is fixed in the PR. Canceled context can lead to failures.
- ping
  - ping does not have this issue as the inner probe runner does not respect context closure
- tcp
  - tcp uses the `sched` utility to run probes, which is susceptible to this failure. `sched` has been updated in the same way as `http`
- udp
  - udp doesn't have the issue because context cancellation will immediately stop the function that does metrics exporting
- udp listener
  - udp listener doesn't have the issue because context cancellation will immediately stop the function that does metrics exporting

### DIFF
--- a/probes/external/external.go
+++ b/probes/external/external.go
@@ -430,7 +430,7 @@ func (p *Probe) runProbe(startCtx context.Context) {
 	if p.mode == "server" {
 		p.runServerProbe(probeCtx, startCtx)
 	} else {
-		p.runOnceProbe(probeCtx)
+		p.runOnceProbe(probeCtx, startCtx)
 	}
 }
 
--- a/probes/external/external.go
+++ b/probes/external/external.go
@@ -320,7 +320,7 @@ func (p *Probe) setupStreaming(c *exec.Cmd, target endpoint.Endpoint) error {
 	return nil
 }
 
-func (p *Probe) runOnceProbe(ctx context.Context) {
+func (p *Probe) runOnceProbe(ctx context.Context, startCtx context.Context) {
 	var wg sync.WaitGroup
 
 	for _, target := range p.targets {
@@ -360,6 +360,14 @@ func (p *Probe) runOnceProbe(ctx context.Context) {
 			}
 
 			err := p.runCommand(ctx, c)
+
+			// exit without saving metrics if the probe was interrupted by context closure
+			// while running the command
+			select {
+			case <-startCtx.Done():
+				return
+			default:
+			}
 
 			success := true
 			if err != nil {
--- a/probes/ping/ping.go
+++ b/probes/ping/ping.go
@@ -533,6 +533,14 @@ func (p *Probe) Start(ctx context.Context, dataChan chan *metrics.EventMetrics) 
 		p.l.Debugf("Probe started, runcount %d", p.runCnt)
 		p.runProbe()
 		p.l.Debugf("Probe finished, runcount %d", p.runCnt)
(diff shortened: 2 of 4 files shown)


probes/ping/ping.go
```diff
@@ -533,6 +533,14 @@ func (p *Probe) Start(ctx context.Context, dataChan chan *metrics.EventMetrics)
...
 		p.runProbe()
 		p.l.Debugf("Probe finished, runcount %d", p.runCnt)
+
+		// exit without exporting stats if the probe run was interrupted.
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+
```
manugarg, line 543: Since runProbe is not impacted by context, this is pretty much inconsequential, right? I'd drop ping.go from this change.

probes/external/external.go
```diff
@@ -422,7 +430,7 @@ func (p *Probe) runProbe(startCtx context.Context) {
 	if p.mode == "server" {
 		p.runServerProbe(probeCtx, startCtx)
 	} else {
-		p.runOnceProbe(probeCtx)
+		p.runOnceProbe(probeCtx, startCtx)
```
manugarg, line 433: probeCtx is already tied to startCtx, so you can just use probeCtx.

probes/external/external.go
```diff
@@ -361,6 +361,14 @@ func (p *Probe) runOnceProbe(ctx context.Context) {
 
 			err := p.runCommand(ctx, c)
 
+			// exit without saving metrics if the probe was interrupted by context closure
+			// while running the command
+			select {
+			case <-startCtx.Done():
```
manugarg, line 367: As commented below, you can probably just use ctx here and skip passing startCtx to this method.

manugarg:
@AdamEAnderson Thanks for this PR. I think it makes sense. I added a few comments.

manugarg:
Hi @AdamEAnderson, just wanted to check if you want to follow up on this PR.

kathleenmotley49-maker (approved):

manugarg:
Unfortunately this PR is stale and a lot has changed since the PR. I still like the idea. I am changing the scheduler in this PR: https://github.com/cloudprober/cloudprober/pull/1147. HTTP probe fully uses scheduler now so that should get automatically fixed. I don't think external probe is impacted as much, because we don't refresh target independently -- each probe loop runs with a given set of targets.

PR #819, opened 2024-08-06
[metrics] Remove redundant cloning of some metrics

manugarg:
Cloning of complex values like latency and response codes is unnecessary now as EventMetrics are cloned right after they are composed[1].

[1] Cloning right after creation sounds bad, but cloning of EventMetrics is handled in a central function (providing more safety and better handling), while creation of EventMetrics happens in many places.
https://github.com/cloudprober/cloudprober/blob/8eabdeba7c2f94135b12cfc48c3117e034eac5fc/probes/options/options.go#L292


manugarg:
I am still thinking about this. Maybe it will be cleaner to provide an option for RecordMetrics to ask for cloning, and keep cloning where it is today.

manugarg:
#820 is a better option I think.

PR #825, opened 2024-08-15
[targets] Provide a way to use cached IP on resolve error

manugarg:


manugarg:
We should see if providing an option to control local DNS TTL is a better way to do this.

manugarg:
#831 is probably a better solution for this.

manugarg:
Closing this as #831 is merged in now.

Issue #826, opened 2024-08-15
Support file path globs in config include


rhodesn:
**Describe the feature you'd like and the problem it will solve**
It would be great if the new config `include: "xxx"` supported file path globs. It's fairly standard practice to have a drop-in style directory, e.g. `/etc/cloudprober/conf.d/*.cfg`, and maintaining a hardcoded list of files that live in this directory could quickly become cumbersome.

manugarg:
It will go out in the next release, which is due next week.

PR #827, opened 2024-08-15
Support file glob when including config files

rhodesn:
This is a potential solution for #826 to enable providing a file path glob in the `include` config to pick up all files in a directory.
I also toyed with the idea of allowing absolute paths in the include config as well but left it out to keep PR in sync with the issue/summary. Something along the lines of:
```
               var absPath string
               if filepath.IsAbs(m[1]) {
                       absPath = m[1]
               } else {
                       absPath = filepath.Join(baseDir, m[1])
               }

               files, err := filepath.Glob(absPath)
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,11 +106,23 @@ func handleIncludes(baseDir string, content []byte) (string, error) {
 			final = append(final, line)
 			continue
 		}
-		includedCfg, err := readConfigFile(filepath.Join(baseDir, m[1]))
+
+		includePath := filepath.Join(baseDir, m[1])
+		files, err := filepath.Glob(includePath)
 		if err != nil {
 			return "", err
 		}
-		final = append(final, includedCfg)
+		if len(files) == 0 {
+			return "", fmt.Errorf("config include error: %s didn't match any files", includePath)
+		}
+
+		for _, file := range files {
+			includedCfg, err := readConfigFile(file)
+			if err != nil {
+				return "", err
+			}
+			final = append(final, includedCfg)
+		}
 	}
 
 	newline := "\n"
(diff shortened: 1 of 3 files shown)


manugarg:
@rhodesn Thanks for the PR. Just wanted to give you a heads up that I've looked at it, but not reviewed it fully yet. I'll try to review it this week.

rhodesn:
If this does make it in I can send a follow up for the docs.

config/config.go
```diff
@@ -106,11 +106,22 @@ func handleIncludes(baseDir string, content []byte) (string, error) {
...
 		}
-		includedCfg, err := readConfigFile(filepath.Join(baseDir, m[1]))
+
+		files, err := filepath.Glob(filepath.Join(baseDir, m[1]))
 		if err != nil {
 			return "", err
 		}
-		final = append(final, includedCfg)
+		if len(files) == 0 {
+			return "", errors.New("config include missing file")
```
manugarg, line 115: I think we can provide more information in the error message:

```suggestion
			return "", fmt.Errorf("config include error: %s didn't match any files", m[1])
```

config/config.go
```diff
@@ -106,11 +106,22 @@ func handleIncludes(baseDir string, content []byte) (string, error) {
 			final = append(final, line)
 			continue
 		}
-		includedCfg, err := readConfigFile(filepath.Join(baseDir, m[1]))
+
+		files, err := filepath.Glob(filepath.Join(baseDir, m[1]))
```
manugarg, line 110:
```suggestion
                 includePath := filepath.Join(baseDir, m[1])
		files, err := filepath.Glob(includePath)
```

manugarg (approved):
I've a couple of super minor comments. Looks good otherwise.

rhodesn:
Thanks @manugarg, both comments addressed.

manugarg (approved):

PR #840, opened 2024-08-30
[probes.http] Migrate HTTP probe partially to common/sched

manugarg:
This is to reduce duplication in the scheduling code. Our. common scheduler code was anyway based on HTTP probe.

Note: there is still some duplication in the startForTarget functionality, but it's the right balance right now without making it too complicated. I'll keep thinking of ways to reduce the duplication further.

### DIFF
--- a/probes/http/http.go
+++ b/probes/http/http.go
@@ -562,7 +562,7 @@ func (p *Probe) startForTarget(ctx context.Context, target endpoint.Endpoint, da
 	clients := p.clientsForTarget(target)
 	for ts := time.Now(); true; ts = <-ticker.C {
 		// Don't run another probe if context is canceled already.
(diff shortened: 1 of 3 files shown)


probes/http/http.go
```diff
@@ -598,138 +593,14 @@ func (p *Probe) startForTarget(ctx context.Context, target endpoint.Endpoint, da
 	}
 }
 
-func (p *Probe) gapBetweenTargets() time.Duration {
-	interTargetGap := time.Duration(p.c.GetIntervalBetweenTargetsMsec()) * time.Millisecond
-
```
manugarg, line 603: This part got lost in this PR. Fix is in: https://github.com/cloudprober/cloudprober/pull/886

Issue #356, opened 2023-04-20
Add support for creating metrics from HTTP probe output


manugarg:
I had opened ticket a long time back at the older location:
https://github.com/google/cloudprober/issues/297

Recently while running a probe, I realized it can be quite a useful thing to be able to build metrics from HTTP probe's output.

manugarg:
From the discussion on #508, I realized that we can probably go a step further and use HTTP headers for custom metrics as well. Something like:

output_metric_options {
  header_to_metrics {
    header: LastModified (LastModified | ContentLength)
    metric: last-modified
  }
  header_to_metrics {
    header: ContentLength
    metric: content-length
  }
  ...
}

manugarg:
We can go a step further and provide an option generate metrics from JSON when output is JSON.

manugarg:
This is now down. You can generate metrics from HTTP response headers, JSON, and plain old line based content.

For example, to generate `content_length` metric by adding something like the following to your `http_probe` config stanza:

```
    response_metrics_options {
      header_metric {
        metric_name: "content-length"
        type: INT
      }
    }
```

manugarg:
You can do even crazier stuff with `json_metric`. See examples here:
https://cloudprober.org/docs/config/metrics/#cloudprober_metrics_payload_OutputMetricsOptions

PR #841, opened 2024-09-01
[probes.http] Support creating metrics from HTTP (or any other) response

manugarg:
This is a big change. It revamps how we create metrics from probe output, making it more generic and supporting more formats.

- Add support for creating metrics from HTTP headers. For example, you can create a `content_length` metric with a small config change. You can use this to parse custom headers as well, that your target may export.
- Add support for creating metrics from JSON output. This is very useful as JSON is a commonly used format on web today.


manugarg:
Attempt at implementing #356.

PR #863, opened 2024-09-27
[probes.browser] Add a web server to serve test artifacts.

manugarg:
- Add a web server to serve test artifacts. Web server serves artifacts
  at /artifacts/<probename> path by default. You can change the path
  through config.
- Arrange artifacts by date.
- Add long term storage options, not implemented yet though.


manugarg:
Of course, the web serve can be improved. Some possible improvements:
- Show directories in a table with columns for last modified date and size (or number of files). We can make the table sortable and searchable.
- Once a particular run is selected, we should show a table of all the individual tests and thumbnails associated with them. Thumbnails should be clickable, and so on.

PR #865, opened 2024-09-29
[probes.browser] Add support for storing artifacts in S3.

manugarg:
- Create the abstration of artifacts handler that will be initialized at the time of
   probe initialization. Artifacts handler in turn will initialize various storage backends
   based on the configuration.

- Probe calls handle() on the artifacts handler in a goroutine after each probe run
  with the output directory path. Artifacts handler will then upload the artifacts to
  all the configured storage backends.

- Tested with a local minio setup and AWS S3 bucket.

- Other changes: enable HTML reporter and persist only reports as these reports
   provide a good overview and contain all the assets like screenshots, videos and
   traces.


manugarg:
Example of S3 bucket (tested using minio):
image
image
image

PR #876, opened 2024-10-04
[probes.http] Make latency breakdown more accurate

edgarasg:
Previously latency breakdown was calculated from probe start time, which means that all those metrics were basically accumulative. This change separates latency of each stage.


manugarg (approved):

manugarg:
Thank you, @edgarasg!

PR #903, opened 2024-11-05
Create an example config for configuring GRPC probes

conallob:
Config lifted from https://github.com/cloudprober/cloudprober/issues/307#issuecomment-1691963331


examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,25 @@
+server {
+  type: GRPC
+
+  grpc_server {
+    port: 3142
+    enable_reflection: true
+  }
+}
```
manugarg, line 8: Having this here is confusing as probe is connecting to this server.

Instead to make it complete, you can add `grpc_port: 9314` at the top level.

examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,25 @@
...
+  name: "grpc_generic"
+  type: GRPC
+  targets {
+    # Hit another cloudprober instance running with grpc_server enabled.
+    host_names: "localhost:9314"
+  }
+
+  grpc_probe {
+    method: GENERIC
+      request {
```
manugarg, line 20: With default server and no TLS certs, it will fail without `insecure_transport: true`.

examples/grpc/grpc_example.cfg
conallob: Done

examples/grpc/grpc_example.cfg
conallob: Added, although for completeness, there should probably be some annotations for how to configure TLS.


I forgot to tick "allow edits by maintainers". Care to suggest TLS configuration for this example?

examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,29 @@
...
+
+probe {
+  name: "grpc_generic"
+  type: GRPC
+  targets {
+    # Hit another cloudprober instance running with grpc_server enabled.
+    host_names: "localhost:9314"
+  }
+
+  grpc_port: 9314
```
manugarg, line 18: So this field exists outside of the probe stanza. Cloudprober's gRPC server is a binary level option. That's the reason test is failing right now.

examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,29 @@
...
+
+  grpc_probe {
+    method: GENERIC
+      request {
+        insecure_transport: true
```
manugarg, line 23:
```suggestion
        # Required if connecting to a server without TLS config
        insecure_transport: true
```

examples/grpc/grpc_example.cfg
manugarg: I'd probably just add a comment like below. In this case it's more of a server side config. Where you specify grpc_port, you can add a reference to the following:
https://github.com/cloudprober/cloudprober/blob/390574610d6a6d641ff343aa962b1aab8221f645/config/proto/config.proto#L86

examples/grpc/grpc_example.cfg
manugarg: Looks like it's still here :)

conallob: My loathing of git continues. `grpc_port` is now at the top level, so should I leave the grpc server for completeness, or not?

examples/grpc/grpc_example.cfg
conallob: Fixed

examples/grpc/grpc_example.cfg
manugarg: Sorry this is not super clear but `server` stanzas add additional servers:
https://cloudprober.org/docs/how-to/built-in-servers/

So, I'd just drop this stanza.

```suggestion
```

conallob:
PTAL

examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  grpc_probe {
+    method: GENERIC
+      request {
+        insecure_transport: true
+        
+        call_service_method: "cloudprober.Cloudprober.ListProbes"
+        body: "{}"
+      }
```
manugarg, line 19: Sorry didn't notice this earlier, but insecure_transport is a `grpc_probe` level field.
```suggestion
    insecure_transport: true
    method: GENERIC
    request {
      call_service_method: "cloudprober.Cloudprober.ListProbes"
      body: "{}"
    }
```

examples/grpc/grpc_example.cfg
```diff
@@ -0,0 +1,20 @@
...
+
+  grpc_probe {
+    insecure_transport: true
+    method: GENERIC
+    request {
+      call_service_method: "cloudprober.Cloudprober.ListProbes"
+      body: "{}"
+    }
+    }
+}
```
manugarg, line 20: Minor indentation fix:
```suggestion
    }
  }
}
```

manugarg (approved):
Just waiting for checks to run before merging.

conallob:
Thanks for the review!

PR #923, opened 2024-11-12
update go directive to include minor version

sfc-gh-bxin:
With the patch number, I got a weird "toolchain not available" error (where my workspace has go 1.22.x installed; and this project is asking to upgrade to 1.23, but apparently it doesn't know how...)

More discussion here:  https://github.com/golang/go/issues/62278

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cloudprober/cloudprober
 
-go 1.23
+go 1.23.3
 
 require (
 	cloud.google.com/go/bigquery v1.59.1


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cloudprober/cloudprober
 
-go 1.23
+go 1.23.3
```
manugarg, line 3: Let's try putting "go 1.23.0"? and master go.mod itself, I'll rebase it afterwards. 

Btw, I am getting close to merging it in the master. I wanted to release v0.13.8 first, which is done now.

sfc-gh-bxin: Ack.

PR #924, opened 2024-11-13
[probes.dns] Add ptype="dns" label to all DNS probe metrics by default

rhodesn:
This appears to be set on other probes. Obviously the workaround is `additional_label`, so I'm not sure whether it's actually worth doing this.


manugarg:
This makes sense. Thank you, @rhodesn!

PR #931, opened 2024-11-15
Update go directive to include the patch number

Redexing:
Without the patch number, I hit a "toolchain not available" error from an env where go 1.22 was installed, and this project requires 1.23, and apparently it cannot locate the toolchain.

Similar discussion in: https://github.com/golang/go/issues/62278

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cloudprober/cloudprober
 
-go 1.23
+go 1.23.0
 
 require (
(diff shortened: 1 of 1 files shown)


manugarg:
To explain this further, starting with Go 1.21, `go` command line tool will download the toolchain specified in `go.mod`, if current tool chain is older than the version in go.mod.

So if go.mod says `go 1.23` and you try to run it with go 1.22, it will try to download go1.23 and fail as go1.23 doesn't exist.

PR #953, opened 2024-12-30
Add cloudprober support for TPC universe domain

rishabhgargsde:
Adding support for the [WithUniverseDomain](https://pkg.go.dev/google.golang.org/api/option#WithUniverseDomain) option in GCP API.

The capability to override the universe domain is needed to support all GCP APIs, including: cloud logging, Pubsub and GCE.

Suggested implementation:

Since the logger.go module already supports few flags ([for example](https://github.com/cloudprober/cloudprober/blob/1a3db5251f1402545f22d8a9f589d904c7817bb5/logger/logger.go#L53)) and is imported by multiple other modules, the most straightforward implementation would be to add an additional startup flag to the logger.go and export this flag as a public property.

### DIFF
--- a/internal/rds/gcp/forwarding_rules.go
+++ b/internal/rds/gcp/forwarding_rules.go
@@ -31,6 +31,7 @@ import (
 	"github.com/cloudprober/cloudprober/logger"
 	"golang.org/x/oauth2/google"
 	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/option"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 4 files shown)


internal/rds/gcp/forwarding_rules.go
```diff
@@ -186,7 +187,13 @@ func defaultComputeService(apiVersion string, baseAPIPath string) (*compute.Serv
...
 		return nil, err
 	}
-	cs, err := compute.New(client)
+
+	opts := []option.ClientOption{option.WithHTTPClient(client)}
+	if *logger.GCPUniverseDomain != "" {
+		opts = append(opts, option.WithUniverseDomain(*logger.GCPUniverseDomain))
+	}
+	var cs *compute.Service
+	cs, err = compute.NewService(context.Background(), opts...)
```
manugarg, line 196: Using logger's fields to determine API options is pretty convoluted. I'd suggest that you use environment variable all across, i.e. if CLOUDPROBER_GCP_UNIVERSE_DOMAIN is set, set the WithUniverseDomain option.

Do it everywhere. Drop the flag support.

rishabhgargsde: Hello @manugarg
Sorry for the late reply. Since our probers already utilize this code and TPC handover is about to happen, we cannot afford any disruptions. Is it possible to sync the code for now and make the changes later?

PR #954, opened 2024-12-30
Support cloudprober container parameters for TPC

rishabhgargsde:
Add parameters to the cloudprober container in accordance with TPC universe ID for migration to artifact registry.


manugarg (commented):
Can you just override this file in the internal repository? This is supposed to be just an example, not supposed to all requirement anywhere.

rishabhgargsde:
Hello @manugarg 
In this case, pulling the changes from the external repository into google3 would end up overwriting our changes. We would have to make sure that this overwriting does not happen each time, manually.

manugarg:
There is a way to skip a file while importing third_party code to google3.

Issue #947, opened 2024-12-18
move probes/external/proto/*.proto to another file to fix gazelle builds


antholeole:
currently, [gazelle](https://github.com/bazel-contrib/bazel-gazelle) is _almost_ able to build this project through auto-generated build files.

the one issue that is stopping it working out of the box is the fact that there are two protos in the `probes/external/proto/` folder, which [gazelle does not support](https://github.com/bazel-contrib/bazel-gazelle/issues/7). 

I've solved this in an internal repository by patching this to concat the two protos in that folder, but it would be nice if this was supported in-trunk so we don't have to maintain a patch. I won't post the patch here because I'm not entirely sure if its the _right_ way to solve this problem, but I'm willing to make the change if we discuss what is the correct direction, if any.

manugarg:
Hello @antholeole! I did some research and it looks like merging config.proto and server.proto won't be a bad idea; see #956 .

> I've solved this in an internal repository by patching this to concat the two protos in that folder, but it would be nice if this was supported in-trunk so we don't have to maintain a patch.

If by internal, you mean Google's internal copy, it's kind of interesting that it ended up being like this because of how it started at Google :)

PR #956, opened 2025-01-03
[probes.external] Merge config and server proto

manugarg:
This makes glaze work for cloudprober, see #947 

There are various side effect of this change:
- Currently Probe{Request,Reply} messages are in `cloudprober` protobuf package but with this move they will be in `cloudprober.probes.external` package. However even with this change, existing clients should continue to work with the new version of cloudprober as package name is not part of the wire format. I verified that using the example included in cloudprober.
- Code generated for Go will continue to work as it is as it will be in the same Go package as earlier.
- For users generating code for their language, they will have to update the code generating scripts/tools to use the merged file (optionally splitting messages - see the proto file for how it works, we are using it for python).
- Other language specific changes:
   - C++ puts all code in the same namespace as proto package. If you've a C++ client you'll need to update it use `cloudprober::probes::external` namespace instead of `cloudprober` namespace.
   - For python, there are no changes in the client. Also, we provide a pypi package for python: https://pypi.org/project/cloudprober-serverutils/


manugarg:
> Nice - thank you!
> 
> By internal repo, we use git on borg, not g3.

I see. 

> Not sure if there's a copy of this in G3, have not checked.

yeah, there is one in third_party.

PR #985, opened 2025-02-01
[build] Download pre-built gotestsum release instead of building it

manugarg:


manugarg:
I am not really sure if saving is worth it. Closing it.

PR #997, opened 2025-02-11
[build] Set caching related GO* variables to "D:" on windows

manugarg:
This is based on the conversation here: https://github.com/actions/setup-go/pull/515


manugarg:
This one seems to have helped. Cache restoration took 27s instead of 3m30s.
image

PR #1003, opened 2025-02-13
[metrics.payload] Allow accepting and rejecting payload lines based on regex

manugarg:
- This will help to avoid cases where there is a spurious output on external probe process failure and we still try to parse it for metrics. For example, this was happening for the browser probe failures -- on failure playwright prompts user to open the report.
- This could in general be a useful feature though.
- [`probes.browser`](diffhunk://#diff-a8df2e33e7b49a165f0786601d2bd8aa4966aa9e07fc5295ebcace41fb227768L238-R241): Use the new feature to only process metrics lines from the reporter.


ericmustin:
Thank you for addressing this @manugarg , I bumped into it. Is there a general release cadence or timeline I can follow? I see v0.14.0 mileston is unreleased, I've been extending the published cloudprober:latest-pw image (which picks up v0.13 currently if i understand correctly) and, while I can point at the cloudprober:main-pw image for now, i'd prefer to use a stable tag.

manugarg:
> Thank you for addressing this @manugarg , I bumped into it. Is there a general release cadence or timeline I can follow? I see v0.14.0 mileston is unreleased, I've been extending the published cloudprober:latest-pw image (which picks up v0.13 currently if i understand correctly) and, while I can point at the cloudprober:main-pw image for now, i'd prefer to use a stable tag.

@ericmustin 
We don't have a published cadence for releases. If there is a critical/wide-impact bug or security issue that impacts Cloudprober, I try to release a fix as early as possible. Other than that, I mostly cut a new release when there are significant changes and they are looking stable -- looking at the past few releases that seems to happen at around 3-4 months mark.

I'll try cut a new release early next week.

By the way, it's cool that you're already using the browser probe type 😄.  I was planning to announce it more broadly once its config API was more stable and there was some documentation. Good news is it's getting there, but there have been some config changes (mostly in artifacts_options) since the last release, not sure if you've caught them already.

manugarg:
@ericmustin Just cut a new release: https://github.com/cloudprober/cloudprober/releases/tag/v0.14.0

PR #1019, opened 2025-02-26
[probes.ping] add new metric for last ping latency

krzee:
This adds a new metric to ping prober named `last_ping_latency`
i don't know go at all, i had help from grok3 making this PR. I apologize if I did this wrong.
I tested it and it does work for me.
The metric probably makes more sense as a gauge instead of a counter. Maybe you have a better name for it.
 Please edit this in any way you desire so that it may be merged.

### DIFF
--- a/probes/ping/ping.go
+++ b/probes/ping/ping.go
@@ -72,6 +72,7 @@ const (
 type result struct {
 	sent, rcvd        int64
 	latency           metrics.LatencyValue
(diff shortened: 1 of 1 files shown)


manugarg:
There is a better way to get per-packet latency:
https://github.com/cloudprober/cloudprober/discussions/121#discussioncomment-12332980

PR #1032, opened 2025-03-06
[probes] Initialize probe config with zero values if config is not given

Koaxiel:
Exactly like for the PR https://github.com/cloudprober/cloudprober/pull/237 but for UDP and External probing

### DIFF
--- a/probes/udp/udp.go
+++ b/probes/udp/udp.go
@@ -139,6 +139,9 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 	}
 	p.src = sysvars.GetVar("hostname")
 	p.c = c
+	if p.c == nil {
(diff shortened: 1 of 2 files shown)


probes/external/external.go
```diff
@@ -119,6 +119,9 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 		p.l = &logger.Logger{}
 	}
 	p.c = c
+	if p.c == nil {
+		p.c = &configpb.ProbeConf{}
+	}
```
manugarg, line 124: Zero config doesn't work for external probe at all. We should return an error if an external probe config is nil.

Koaxiel: You are right, as `command` is required.
I remove my change on external as it's already handle via `GetCommand()`

PR #1035, opened 2025-03-08
[probes.udp] Fix a bug in packets processing

manugarg:
- While processing packets, we were not updating highest sequence number when processing sent packets from the last packet processing cycle, this was causing some received packets to be punted for very long. This resulted in zigzag metrics, where success ratio unnecessarily jumped around 1.
image

- Graph after the fix:
image


manugarg:
@ls692 You'd remember this issue -- where cloudradar dashboard will sometimes show negative packet loss :)

Issue #958, opened 2025-01-04
[probes.browser] Consider providing more advanced artifacts web server UI


manugarg:

manugarg:
More details:

<img width="1075" alt="Image" src="https://github.com/user-attachments/assets/6ce00905-79ab-4941-93c6-8074bc468d55" />

manugarg:
With #1074 and #1077 things are much nicer now. I am thinking of skipping the pagination (it never looks nice) -- instead one should just rely on adjusting the time range.

Thinking of couple of other changes:
- Collapse output at date by default if too many entries.
- Show time corresponding to the timestamps.

manugarg:
I think we'll stop here on this one. New UI looks pretty good:

![Image](https://github.com/user-attachments/assets/3feb6845-9164-4a80-92ba-8d9f3b771fce)

PR #1074, opened 2025-05-01
[probes.browser] First check-in for advanced web version

manugarg:
This still needs a lot of work for the full implementation, but this version is already quite useful as it lists all the timestamp directories on the top page (see screenshot in the comment below). It'll now be easy to add other features like filtering by time range.

One drawback of this current design is that it shows all timestamped directories on one page (instead of showing them by days), which may be hard to browse. I'll check this in and follow-up with filtering by date and time in the next PR.

See https://github.com/cloudprober/cloudprober/issues/958 for more context.


manugarg:
image

PR #1086, opened 2025-05-28
let probes running with in-tree scheduler run with different context from probe lifecycle context

Redexing:
when a probe target endpoints change, probes for obsolete endpoints are stopped by the scheduler, by calling cancelFunc on its lifecycle context;   for each probe run, its context is chained from the same lifecycle context, leading to such in-progress probes being cancelled (and counted as failures).

Use a different context, but still with the intended timeout, for the probe run to fix this issue.

Alternative solution: make sure that target refresh schedule and probe run schedule are "well spaced/separated" from each other; but this becomes trick as probe frequency increases and number of targets increases.


kathleenmotley49-maker (approved):

manugarg:
Hi Bin,

Finally, got a chance to review this in more detail. I think the overall idea makes sense. But I'd like to change the implementation a bit so that we at least use the probe context instead of background context.

1. Rename `probeCtx` here to  `probeTargetCtx`:
https://github.com/cloudprober/cloudprober/blob/403ab84188cf9e32594cb6ade6899602cb8f5c5e/probes/common/sched/sched.go#L231

2. Pass both probe level context `ctx` and target level ctx `probeTargetCtx` to `startProbeForTarget`. In [`startProbeForTarget`](https://github.com/cloudprober/cloudprober/blob/403ab84188cf9e32594cb6ade6899602cb8f5c5e/probes/common/sched/sched.go#L128) take target level ctx as simply `ctx` and probe-level ctx as `probeCtx`.

3. At line 157 (current change), have it use `ctx` or `probeCtx` based on a Scheduler option. Scheduler can have an additional field called `UseProbeLevelContext` -- in the comment add an explanation how is it different.

What do you think?

manugarg:
I think it still makes sense to do this, but https://github.com/cloudprober/cloudprober/pull/1147 might be a sufficient solution for the issues we are seeing. So it's not that urgent anymore.

kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

PR #1105, opened 2025-07-22
Use net.JoinHostPort to combine host and port

kaoet:
When the host is an IPv6 address like `::1`, simple `fmt.Sprintf()` doesn't surrround the host with brackets, while `net.JoinHostPort` produces correct result like `[::1]:80`.


manugarg (approved):
LGTM. Thanks!

PR #1147, opened 2025-09-18
[probes.sched] Skip exporting metrics if context is canceled

manugarg:


probes/common/sched/sched.go
```diff
@@ -154,12 +154,15 @@ func (s *Scheduler) startForTarget(ctx context.Context, target endpoint.Endpoint
...
-		if (runCnt % s.Opts.StatsExportFrequency()) == 0 {
+		timedCtx, cancelTimedCtx := context.WithTimeout(ctx, s.Opts.Timeout)
+		s.RunProbeForTarget(timedCtx, runReq)
+		cancelTimedCtx()
+
+		// Export stats if it's the time to do so and context was not canceled
+		// while we were running the probe. Context is typically canceled when
+		// target is deleted after a target refresh. We don't want to export
+		// metrics in such cases.
+		if (runCnt%s.Opts.StatsExportFrequency()) == 0 && !CtxDone(ctx) {
```
Redexing, line 165: Might be better to explicitly test for "cancelled" condition (via `Context#Err` test), to avoid dropping metrics when successful probe runs finish _just_ within timeout.   You may need to move the `cancelTimedCtx` to the end (or jus use defer)

manugarg: We chatted offline -- we now use a different context for timed run. This particular context (ctx) will only finish on cancelation.

PR #1158, opened 2025-10-20
[probes.http] Add merge_header_metrics option to HTTP probe

PovilasV1:
##
When using `response_metrics_options` with `header_metric`, cloudprober outputs header metrics on separate lines, making downstream processing difficult as metrics need to be joined by timestamp and labels.

##
Added optional `merge_header_metrics` configuration (default: `false`) to merge header metrics into the main probe output line.

### Configuration
```protobuf
http_probe {
  response_metrics_options {
    header_metric {
      header_name: "server"
      metric_name: "server_header"
    }
  }
  merge_header_metrics: true
}
```

### Output comparison

**Before (or with `merge_header_metrics: false`):**
```
cloudprober ... total=1 success=1 latency=123.456 resp-code=map:code,200:1
cloudprober ... server_header="LiteSpeed"
```

**After (with `merge_header_metrics: true`):**
```
cloudprober ... total=1 success=1 latency=123.456 resp-code=map:code,200:1 server_header="LiteSpeed"
```

### DIFF
--- a/probes/http/http.go
+++ b/probes/http/http.go
@@ -87,6 +87,7 @@ type probeResult struct {
 	latencyBreakdown             *latencyDetails
 	sslEarliestExpirationSeconds int64
 	payloadMetrics               []*metrics.EventMetrics
+	mergeHeaderMetrics           bool
 }
 
 func (p *Probe) getTransport() (*http.Transport, error) {
@@ -423,6 +424,7 @@ func (p *Probe) newResult() *probeResult {
 	result := &probeResult{
 		respCodes:                    metrics.NewMap("code"),
 		sslEarliestExpirationSeconds: -1,
+		mergeHeaderMetrics:           p.c.GetMergeHeaderMetrics(),
 	}
 
 	if p.c.GetKeepAlive() {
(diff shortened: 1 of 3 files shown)


manugarg:
@PovilasV1 Unfortunately merging metrics like this is problematic as both EventMetrics have different types (gauge vs cumulative).

What problem are you trying to solve? Most surfacers don't care about metrics being on different lines (different EventMetrics). Which surfacer are you using?

PovilasV1:
We’re not using a surfacer. Our pipeline works as follows: Cloudprober prints probe results to stdout, rsyslog captures them to /var/log/messages, Fluentd monitors and parses the logs and then forwards the processed data to BigQuery. In that case we will probably look into changing the pipeline and start using a surfacer.

PovilasV1:
Thanks! I think we will try to adopt the BigQuery surfacer, but we need to figure out how to keep the same structure. Will try to look into this https://github.com/cloudprober/cloudprober/discussions/1076#discussioncomment-13136905

manugarg:
Ack. IIRC, even with bigquery surfacer, each metric goes to a different row. But having the single timestamp will allow you to group-by more easily.

PR #1166, opened 2025-10-31
[build] release xds variant binaries

otkd:
Since xds was excluded by default in #467 added them as separate precompiled release

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -145,6 +145,12 @@ jobs:
       - run: make cloudprober-macos-{amd64,arm64} cloudprober-windows-amd64
         if: ${{ matrix.os == 'macos-latest' }}
 
+      - run: BUILD_WITH_XDS=true make cloudprober-xds-linux-{amd64,arm64,armv7}
+        if: ${{ matrix.os == 'ubuntu-latest' && startsWith(github.ref, 'refs/tags/v') }}
+
+      - run: BUILD_WITH_XDS=true make cloudprober-xds-macos-{amd64,arm64} cloudprober-xds-windows-amd64
+        if: ${{ matrix.os == 'macos-latest' && startsWith(github.ref, 'refs/tags/v') }}
+
       - name: Upload cloudprober dist
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -41,10 +42,19 @@ cloudprober-fips: $(SOURCES)
 	go tool nm cloudprober-fips | grep crypto/internal/boring/sig.BoringCrypto.abi0 > /dev/null || (echo "FIPS build failed: BoringCrypto not used" && rm cloudprober-fips && exit 1)
 	strip cloudprober-fips
 
+define make-binary-target-xds
+$1: $(SOURCES)
+	GOOS=$(subst macos,darwin,$(word 3,$(subst -, ,$1))) ; \
+	GOARCH=$(subst armv7,arm,$(word 4,$(subst -, ,$1))) ; \
+	GOARM=$(subst armv7,7,$(filter armv7,$(word 4,$(subst -, ,$1)))) ; \
+	CGO_ENABLED=0 GOOS=$$$${GOOS} GOARCH=$$$${GOARCH} GOARM=$$$${GOARM} go build -tags grpc_xds -o $1 -ldflags $(LDFLAGS) $(BINARY_SOURCE)
+endef
```
manugarg: I won't duplicate this and the above stuff. Instead:

Define a couple of variables:
BINARY_PREFIX ?= cloudprober
ADDITIONAL_BUILD_TAGS := ''

and, use these variables in the existing place, e.g.
```
BINARIES := $(addprefix $(BINARY_PREFIX)-, $(LINUX_PLATFORMS) macos-amd64 macos-arm64 windows-amd64)
```

Makefile
```diff
@@ -41,10 +42,19 @@ cloudprober-fips: $(SOURCES)
...
+
 test:
 	go test -v -race -covermode=atomic ./...
 
 $(foreach bin,$(BINARIES),$(eval $(call make-binary-target,$(bin))))
+$(foreach bin,$(BINARIES_XDS),$(eval $(call make-binary-target-xds,$(bin))))
```
manugarg: You won't need separate target for xds binaries -- command line would determine which targets will be defined.
```suggestion
$(foreach bin,$(BINARIES),$(eval $(call make-binary-target,$(bin))))
```

Makefile
```diff
@@ -74,6 +84,17 @@ dist: $(BINARIES)
...
+	for bin in $(BINARIES_XDS) ; do \
+	  bindir=$${bin/amd64/x86_64}; \
+	  bindir=$${bindir/cloudprober-xds-/}; \
+	  bindir=cloudprober-xds-$(VERSION)-$${bindir}; \
+	  mkdir -p $${bindir}; cp $${bin} $${bindir}/cloudprober; \
+	  chmod a+rx $${bindir}/cloudprober; \
+	  [[ "$${bin}" == *"windows"* ]] && mv $${bindir}/cloudprober{,.exe}; \
+	  zip -r $${bindir}.zip $${bindir}/; rm -rf $${bindir}; \
+	done
+
```
manugarg: Same as above calling:
BUILD_WITH_XDS=true make dist

should do the right thing, just replace cloudprober- with $(BINARY_BASE)- in dist rule

.github/workflows/go.yml
```diff
@@ -192,6 +192,47 @@ jobs:
...
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - name: Check out code into the Go module directory
+        if: ${{ !contains(github.event_name, 'workflow_dispatch') }}
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check out code into the Go module directory
```
manugarg: Please don't add another workflow. Maintaining all these different workflows is already some work. Just build xds binaries with the above workflow, they will go to the same artifact.

manugarg (commented):
There seem to be a lot of avoidable duplication in this PR. Comments inline.

Makefile
```diff
@@ -28,10 +40,15 @@ endif
...
-	CGO_ENABLED=0 GOOS=$$$${GOOS} GOARCH=$$$${GOARCH} GOARM=$$$${GOARM} go build -o $1 -ldflags $(LDFLAGS) $(BINARY_SOURCE)
+	BINNAME=$1; \
+	BINNAME=$$$${BINNAME#$(BINARY_BASE)-}; \
+	GOOS=$$$$(echo $$$${BINNAME} | cut -d- -f1); \
+	GOOS=$$$${GOOS/macos/darwin}; \
+	GOARCH=$$$$(echo $$$${BINNAME} | cut -d- -f2); \
+	GOARCH=$$$${GOARCH/armv7/arm}; \
+	GOARM=; \
+	[[ "$$$${BINNAME}" == *"armv7"* ]] && GOARM=7; \
+	CGO_ENABLED=0 GOOS=$$$${GOOS} GOARCH=$$$${GOARCH} GOARM=$$$${GOARM} go build $(BUILD_TAG_FLAG) -o $1 -ldflags $(LDFLAGS) $(BINARY_SOURCE)
```
manugarg: Do we need to really change this. I think just adding $(BUILD_TAG_FLAG) will be sufficient. This has worked so well until now that I'd rather not change it if not needed.

Issue #1063, opened 2025-04-15
Publish Cloudprober into Consul Service Discovery


conallob:
Explicitly scoping https://github.com/cloudprober/cloudprober/issues/864 to querying Consul for querying Consul for service discovery.

But to compliment querying Consul, it should also be possible to configure Cloudprober, either via configuration or CLI flags to publish itself to a Consul instance to support service discovery.

https://developer.hashicorp.com/consul/docs/connect/native/go appears to be the best docs for how to implement native consul publishing. It is labelled as deprecated though, but https://github.com/hashicorp/consul/issues/10339 has been open since 2021, so I don't believe xDS is going to launch in the immediate future

Issue #864, opened 2024-09-27
Support Consul Target Discovery


conallob:
To facilitate integration with network management tooling, it would be nice if CloudProber could be integrated with Hashicorp Consul to dynamically configure targets.

This could be implemented as an RDS, but since the use case would be beneficial to others, perhaps adding it as a native target type would be more beneficial.

manugarg:
Funny thing, I recently closed an old issue for the same because of lack of demand: 😄 
 https://github.com/cloudprober/cloudprober/issues/8

manugarg:
Just wanted to add that consul targets implementation should still happen through RDS -- basically add a new RDS provider similar to `file`, `kubernetes`, and `gcp`. Implementation will look similar to this PR: https://github.com/cloudprober/cloudprober/pull/640 (unfortunately this PR got abandoned, but approach was correct).

conallob:
Thanks for the pointers, I am going to work on implementing this

Issue #8, opened 2021-10-29
Add dynamic discovery support for consul catalog based targets


manugarg:
_Copied from https://github.com/google/cloudprober/issues/316, originally filed by @manugarg._

Consul seems to be a popular option for service registry and discovery. Adding dynamic discovery support for consul based targets to cloudprober may be useful to users across cloud platforms.

Targets config for consul will probably look something like this:

```
probe {
 ...
 targets {
   rds_targets {
    resource_path: "consul://service/my-service" (or /nodes)
    filter {
      key: "name"
      value: "{{.regex}}"
    }
  }
  ...
}
```

manugarg:
Closing, as there has not been any real demand for this.

PR #1192, opened 2025-12-26
[targets.consul] Add Consul RDS for Targets

conallob:
Implements Consul as an RDS mechanism, querying a Consul instance to dynamically identify targets via Consul Service Discovery:
- Support both the `targets { consul { ... } }` shorthand and more verbose `rds_targets` configurations
- Register Cloudprober itself with Consul via CLI flags (`--consul_registration`, `--consul_addr`, etc.)

Implements https://github.com/cloudprober/cloudprober/issues/1063 , https://github.com/cloudprober/cloudprober/issues/864 and https://github.com/cloudprober/cloudprober/issues/8

## Design

- **Consul RDS** (`internal/rds/consul/`): Uses the Consul HTTP REST API directly via `net/http` — no dependency on `github.com/hashicorp/consul/api`. Follows the same pattern as the Kubernetes RDS provider. Supports filtering by service name, tags, health status, and metadata. Propagates Consul metadata as labels for use with `@target.label.KEY@`.
- **Self-registration** (`internal/registration/consul/`): A CLI-flag-driven module (not a surfacer) that registers the Cloudprober instance with Consul on startup and deregisters on shutdown. Uses the Consul HTTP REST API directly.

## Co-Authored-By

Claude Code + Claude Sonnet 4.6, with the prompts:

```
Subclass https://github.com/cloudprober/cloudprober/tree/main/internal/rds and
add a new service discovery mechanism, Hashicorp's Consul. It should be possible to
define a Consul service globally, or per probe definition, either by hostname:port or via
Kubernetes service discovery. Probe definitions and filter inputs should be integrated
with how to query a Consul instance.
```

```
Remove @internal/surfacers/consul/consul.go, moving the logic to publish existence
and health state into a golang module, configured via CLI flags.
Then integrate the feedback from @manugarg on https://github.com/cloudprober/cloudprober/pull/1192
```


conallob:
@manugarg This is now ready for your review, but I am not able to set you as the PR assignee

manugarg (commented):
In general, no. As I said on Slack, these kind of PRs don't work. It's un-reviewable.

Other couple of things:
- Claude / you got what surfacers mean completely wrong. Surfacers surface metrics not Cloudprober's own discovery.
- I'd like to avoid taking dependency on Hashicorp APIs just for this RDS. You can use HTTP REST APIs similar to what we do for kubernetes.

PR #1195, opened 2026-01-02
[servers.http] Custom Response headers from HTTP server

### LINKED ISSUE #890: Custom Response headers from HTTP server
I'm proposing that we add an extra field to https://github.com/cloudprober/cloudprober/blob/master/internal/servers/http/proto/config.proto to allow servers to send custom headers in responses. Does this sound reasonable?

majiayu000:
Fixes #890

## Changes
- Add response_header map field to HTTP server proto config
- Implement header injection in HTTP server handler
- Add test for custom response headers functionality

### DIFF
--- a/internal/servers/http/http_test.go
+++ b/internal/servers/http/http_test.go
@@ -198,8 +198,9 @@ func TestCustomResponseHeaders(t *testing.T) {
 
 	dataChan := make(chan *metrics.EventMetrics, 10)
 	s := &Server{
-		c:             c,
-		ln:            ln,
+		c:         c,
+		ln:        ln,
+		reqMetric: metrics.NewMap("url"),
 		staticURLResTable: map[string][]byte{
 			"/": []byte(OK),
 		},


internal/servers/http/http_test.go
```diff
@@ -175,3 +177,54 @@ func TestLameduckListerNil(t *testing.T) {
...
+		ln:            ln,
+		statsInterval: 2 * time.Second,
+		instanceName:  "testInstance",
+		ldLister:      &fakeLameduckLister{},
+		reqMetric:     metrics.NewMap("url"),
+		sysVars:       map[string]string{"instance": "testInstance"},
+		staticURLResTable: map[string][]byte{
+			"/": []byte(OK),
+		},
+	}
```
manugarg: You can leave out setting all these. Just set c, ln, and the staticURLResTable
```suggestion
	s := &Server{
		c:             c,
		ln:            ln,
		staticURLResTable: map[string][]byte{
			"/": []byte(OK),
		},
	}
```

manugarg (commented):
LGTM. One minor suggestion.

PR #1199, opened 2026-01-15
[examples.percentiles] Annotate and Improve Findability of How to Configure Percentiles

conallob:
* Annotate the scope of the percentiles example in https://cloudprober.org/docs/how-to/percentiles/ , as a reminder for 
  readers to not blindly C&P the example

* Create an explicit example config under `examples/percentiles/`

### DIFF
--- a/docs/content/docs/how-to/percentiles/index.md
+++ b/docs/content/docs/how-to/percentiles/index.md
@@ -33,6 +33,7 @@ probe {
     host_names: "..."
   }
 
+  # Local and Intra-regional latencies
   latency_unit: "ms"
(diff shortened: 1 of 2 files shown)


manugarg (commented):
Unreviewable because of crazy diffs. But if we keep them aside, I am wondering what's the use of such PRs. We can't create example for everything and I've not seen anybody run into these problems before. There is a whole doc on configuring percentiles: https://cloudprober.org/docs/how-to/percentiles/

conallob:
Friendly ping?

The diff is now +14 lines following the rebae

PR #1216, opened 2026-02-08
[probes.http] Fix the `max_redirects` handling

rweir:
Hi!  I found that I just couldn't get cloudprober to successfully validate a `301` while probing a redirect that I'd set up, no matter what I did with `max_redirects:`.  After some more digging, I think that PR #1081 [dropped](https://github.com/cloudprober/cloudprober/pull/1081/changes#diff-3f5c32add43466b0c688a1be7fd499fc19d45a1cb7dcdd7c357b241cdc194a0bL542) the line that installed the `redirectFunc` on the `http.Client` that is used for probing, so now it's always implicitly `nil`, so it will accept up to whatever the default redirect counter is for the Go http client lib (10, I think).  This PR has a slightly awkward test demonstrating the issue and a two line patch to fix it.

[bug report](https://github.com/cloudprober/cloudprober/issues/1217).

### DIFF
--- a/probes/http/http.go
+++ b/probes/http/http.go
@@ -539,9 +539,9 @@ func (p *Probe) httpClient(target endpoint.Endpoint) *http.Client {
 			}
 		}
 
-		return &http.Client{Transport: t}
+		return &http.Client{Transport: t, CheckRedirect: p.redirectFunc}
 	}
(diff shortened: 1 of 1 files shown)


rweir:
No trouble at all, and thanks for `cloudprober`!

PR #1255, opened 2026-03-16
[probes.sched] Added custom interval for RDS discovery call

arshbhatti8:
Add a new `targets_update_sec` field to the `ProbeDef` proto, allowing users to configure how often the probe scheduler re-evaluates its target list on a per-probe basis.

Currently Cloudprober's probe scheduler re-evaluates (discovers) targets on a fixed 1-minute interval. For dynamic infrastructure — such as RDS-discovered targets in environments where instances spin up and down frequently — this default is too slow to detect changes promptly. There was no way to configure this interval per-probe, forcing users to either tolerate stale target lists or work around it at a higher level.

### Changes:

- **probes/proto/config.proto** — Adds `optional int32 targets_update_sec = 103` to `ProbeDef`.
- **probes/options/options.go** — Reads `targets_update_sec` from the probe config and sets `TargetsUpdateInterval` in `Options`.
- **probes/common/sched/sched.go** — Applies `TargetsUpdateInterval` from options when initializing the scheduler, overriding the default before the existing floor check against the probe interval.
- **probes/common/sched/sched_test.go** — Adds tests for the scheduler init logic covering default, custom, clamped, and zero-value cases.

## Example config
```
probe {
  name: "my_probe"
  type: HTTP
  interval: "5s"
  timeout: "2s"

  targets {
    rds_targets { ... }
  }

  targets_update_sec: 10  # Re-discover targets every 10 seconds instead of 60
}
```

The existing behavior is preserved — if `targets_update_sec` is not set, the scheduler continues to default to 1 minute, and the floor of `max(targets_update_sec, probe_interval)` is still enforced.

### DIFF
--- a/probes/common/sched/sched_test.go
+++ b/probes/common/sched/sched_test.go
@@ -331,6 +331,58 @@ func TestLastRunResultSet(t *testing.T) {
 	}
 }
 
+func TestSchedulerInit(t *testing.T) {
+	tests := []struct {
+		name                  string
+		probeInterval         time.Duration
+		targetsUpdateInterval time.Duration
+		want                  time.Duration
+	}{
+		{
+			name:          "default when not set",
+			probeInterval: 10 * time.Second,
+			want:          DefaultTargetsUpdateInterval,
+		},
+		{
+			name:                  "custom interval is used",
+			probeInterval:         10 * time.Second,
+			targetsUpdateInterval: 30 * time.Second,
+			want:                  30 * time.Second,
+		},
+		{
+			name:                  "custom interval smaller than probe interval is clamped to probe interval",
+			probeInterval:         2 * time.Minute,
+			targetsUpdateInterval: 10 * time.Second,
+			want:                  2 * time.Minute,
+		},
+		{
+			name:                  "zero custom interval falls back to default",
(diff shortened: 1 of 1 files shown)


manugarg (commented):
Added a couple of comments. Can you also add unit tests for these changes please.

probes/proto/config.proto
```diff
@@ -194,6 +194,12 @@ message ProbeDef {
   // everywhere but run this probe only on a subset of machines.
   optional string run_on = 3;
 
+  // How often to re-evaluate (discover) targets in seconds. This controls the
+  // probe scheduler's target update interval — how often new targets start
+  // being probed and removed targets stop being probed. If not set, defaults
+  // to 1 minute.
+  optional int32 re_eval_sec = 103;
```
manugarg: As I said earlier, this is not the right name. Call it `targets_update_sec`?

probes/options/options_test.go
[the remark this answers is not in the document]
arshbhatti8: Removed. Added test for scheduler changes

probes/proto/config.proto
arshbhatti8: updated the name

PR #1274, opened 2026-03-19
[rds] Expose re_eval_sec consistently across all rds clients ; Expose k8s namespace metadata

conallob:
* Plumb through the ability to set `re_eval_sec` on any rds mechanism consistently, not just for a subset of them

* Include the `namespace` metadata in k8s service discovery results


manugarg:
I think this PR anyway pulled in many unrelated changes and some suggestions of it have been implemented through other PRs in a slightly different way:
- Target update interval -- implemented in a way that it doesn't confuse it with re_eval_sec which is a different mechanism.
- Namespace export -- implemented as a label. Prefixing it to the resource name like in this PR would've broken existing users.

conallob:
Agreed, https://github.com/cloudprober/cloudprober/pull/1255 was a related PR that made this PR redundant

PR #1283, opened 2026-04-03
[surfacers.bigquery] Add wide_rows mode for one-row-per-EventMetrics output

PovilasV1:
Adds a new wide_rows option to the BigQuery surfacer that writes one row per EventMetrics instead of one row per metric. In this mode each metric gets its own column, named via metric_name in bigquery_columns, rather
than being fanned out into generic metric_name/metric_value rows.

Map metrics expand into multiple columns using the map key as a suffix (e.g. resp_code_200), and distribution metrics are written as _sum and _count columns.

Resolves the use-case raised in https://github.com/cloudprober/cloudprober/discussions/1076.


PovilasV1:
> Hi @PovilasV1, can you please add a little bit of more description for this change?

Hi, updated the description. Sorry about that

internal/surfacers/bigquery/bigquery.go
```diff
@@ -183,6 +187,75 @@ func (s *Surfacer) bqLabels(em *metrics.EventMetrics) (map[string]bigquery.Value
...
+		col, hasCfg := metricCols[metricName]
+		if !hasCfg {
+			// Only write metrics explicitly mapped in bigquery_columns.
+			continue
+		}
+		val := em.Metric(metricName)
+		colName := col.GetColumnName()
+
+		// Map metric — write one column per key: <column_name>_<map_key>.
+		if mapVal, ok := val.(*metrics.Map[int64]); ok {
```
manugarg, line 219: This doesn't handle Map[float64]. Make it handle both similar to `parseMapToBQCols[T int64 | float64]`

internal/surfacers/bigquery/bigquery.go
```diff
@@ -183,6 +187,75 @@ func (s *Surfacer) bqLabels(em *metrics.EventMetrics) (map[string]bigquery.Value
...
+
+	row[s.c.GetMetricTimeColName()] = em.Timestamp
+
+	// Build metric_name → BQColumn map for O(1) lookup.
+	metricCols := make(map[string]*configpb.BQColumn)
+	for _, col := range s.c.GetBigqueryColumns() {
+		if col.GetMetricName() != "" {
+			metricCols[col.GetMetricName()] = col
+		}
+	}
```
manugarg: This is static config, we can just build it once in init().

internal/surfacers/bigquery/bigquery.go
```diff
@@ -296,6 +373,21 @@ func (s *Surfacer) writeToBQ(ctx context.Context, inserter iInserter) {
 func (s *Surfacer) init(ctx context.Context) error {
 	s.writeChan = make(chan *metrics.EventMetrics, s.c.GetMetricsBufferSize())
 
+	if s.c.GetWideRows() {
+		hasMetricCols := false
+		for _, col := range s.c.GetBigqueryColumns() {
+			if col.GetLabel() != "" && col.GetMetricName() != "" {
+				s.l.Warningf("bigquery_columns entry %q has both label and metric_name set; label is ignored in wide_rows mode", col.GetColumnName())
```
manugarg: We should return on misconfiguration here. Same for below.

internal/surfacers/bigquery/proto/config.proto
```diff
@@ -37,10 +37,19 @@ message SurfacerConf {
   optional string metric_time_col_name = 9 [default = "metric_time"];
   optional string metric_name_col_name = 10 [default = "metric_name"];
   optional string metric_value_col_name = 11 [default = "metric_value"];
+
+  // If true, write one wide row per EventMetrics instead of one tall row per
+  // metric. Each BQColumn entry must specify a metric_name to select the
+  // metric value written into that column.
+  optional bool wide_rows = 15 [default = false];
 }
```
manugarg: Why skip proto field numbers 12-14?

internal/surfacers/bigquery/bigquery.go
PovilasV1: Should be fixed, now handles both Map[int64] and Map[float64].

internal/surfacers/bigquery/bigquery.go
PovilasV1: Done, both cases now return an error from init() instead of logging a warning.

internal/surfacers/bigquery/proto/config.proto
PovilasV1: No reason to skip them, changed to field 12 and regenerated config.pb.go.

internal/surfacers/bigquery/bigquery.go
PovilasV1: Moved to a metricCols field on Surfacer, built once in init().

PR #1300, opened 2026-05-04
[probes.starlark] Add Starlark script probe (phase 1)

manugarg:
## Summary

Introduces a new `STARLARK` probe type that runs a user-supplied Starlark function once per resolved target each interval. Wall time of the call becomes the probe's latency; clean return is success, any unhandled error (including assertion failures) is failure.

This is **phase 1** — intentionally minimal, anchored on a single canonical use case: *log in, then fetch a protected endpoint with the returned token, assert both succeed.* Anything more is deferred until real demand justifies it.

### Why Starlark

- Hermetic and sandboxed — no surprise filesystem/network access, safe to load from arbitrary config sources.
- No subprocess per run — runs in-process, unlike external probes.
- Single binary, no Python/Node runtime to ship.
- Hot-reloadable; fits cloudprober's "config is data" model.
- Strong precedent: Bazel, Buck2, Tilt, Pants.

### Phase 1 surface

- `http.get(url, headers=)` / `http.post(url, headers=, body=, json=)`
- `Response`: `.status`, `.headers`, `.body`, `.json()`
- `assert.status(response, code)`
- `target.name`, `target.port`

Deliberately **not** in this PR: `dns`/`tcp`/`grpc`/`metric`/`log`/`vars`/`secrets`/`state` modules, latency/body/regex/json_path assertions, follow_redirects, TLS config, structured assertion-failure labels. These come in later phases as real demand surfaces.

### Implementation notes

- Compiled once at `Init`; each run gets a fresh `starlark.Thread` with `ctx.Cancel` wired to the probe's context.
- Builds on `sched.Scheduler` — RunOnce, target discovery, latency metric, and metrics emission work via the standard pipeline.
- The `http` builtin is a thin wrapper over `net/http.Client` — **not** a reuse of `probes/http`. Keeps phase 1 small; revisit if scripts outgrow it.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./probes/starlark/...` passes
- [x] `go test ./probes/starlark/...`
- [x] Manual: run cloudprober with a STARLARK probe at `examples/starlark/...`

🤖 Generated with Claude and Codex


probes/script/builtins.go
```diff
@@ -0,0 +1,325 @@
...
+			k, ok1 := starlark.AsString(item[0])
+			v, ok2 := starlark.AsString(item[1])
+			if !ok1 || !ok2 {
+				return nil, fmt.Errorf("http.%s: headers keys and values must be strings", strings.ToLower(method))
+			}
+			req.Header.Set(k, v)
+		}
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
manugarg: http.DefaultClient is the global default. It follows redirects by default, has no per-script timeout (only the request context), and is shared with anything else in-process that uses http.DefaultClient. The shared client is cheap but the global mutability is a subtle hazard. Consider an explicit &http.Client{} owned by Runtime even if behavior is identical today — costs nothing and removes a shared-state surface.

probes/starlark/builtins.go
```diff
@@ -0,0 +1,325 @@
...
+	defer resp.Body.Close()
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	hdr := starlark.NewDict(len(resp.Header))
+	for k, v := range resp.Header {
+		_ = hdr.SetKey(starlark.String(k), starlark.String(strings.Join(v, ", ")))
+	}
```
manugarg, line 142: Response headers collapse multi-values into a comma-joined string. This is a footgun for Set-Cookie and other repeating headers. Acceptable phase 1 — but worth a one-line comment on response.headers so a future reader doesn't assume HTTP-correct semantics.

PR #1327, opened 2026-05-13
fix: validate trimmed surfacer additional labels

immanuwell:
Fixes a tiny papercut in `CLOUDPROBER_ADDITIONAL_LABELS` parsing.

On `main`, `env=prod, =identity` gets accepted as `[[env prod] [ identity]]`, because the empty-key check runs before trimming. That can leak a blank-ish label key into exported metrics from a normal env var typo.

This change does two things:
- trims key/value before validating, so whitespace-only keys or values are rejected
- splits only on the first `=`, so values like `token=a=b` are preserved correctly

Repro:
```
go test ./surfacers/options -run TestProcessAdditionalLabels
```

Tests:
```
go test ./surfacers/options
go test ./surfacers
```

`go test ./...` hits local ICMP socket permission failures in ping/prober tests here, unrelated to this change.

### DIFF
--- a/surfacers/options/options.go
+++ b/surfacers/options/options.go
@@ -159,12 +159,12 @@ func processAdditionalLabels(envVar string, l *logger.Logger) [][2]string {
 	}
 	var labels [][2]string
 	for _, label := range strings.Split(labelsStr, ",") {
-		kv := strings.Split(label, "=")
-		if len(kv) != 2 || kv[0] == "" || kv[1] == "" {
+		key, val, ok := strings.Cut(label, "=")
+		key, val = strings.TrimSpace(key), strings.TrimSpace(val)
+		if !ok || key == "" || val == "" {
 			l.Warningf("Invalid additional label format: %s", label)
 			continue
 		}
-		key, val := strings.TrimSpace(kv[0]), strings.TrimSpace(kv[1])
 		labels = append(labels, [2]string{key, val})
 	}
 	return labels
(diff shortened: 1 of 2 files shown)


surfacers/options/options.go
```diff
@@ -159,12 +159,16 @@ func processAdditionalLabels(envVar string, l *logger.Logger) [][2]string {
...
+		kv := strings.SplitN(label, "=", 2)
+		if len(kv) != 2 {
 			l.Warningf("Invalid additional label format: %s", label)
 			continue
 		}
 		key, val := strings.TrimSpace(kv[0]), strings.TrimSpace(kv[1])
+		if key == "" || val == "" {
+			l.Warningf("Invalid additional label format: %s", label)
+			continue
+		}
```
manugarg: strings.Cut seems like a better choice for this now:
https://pkg.go.dev/strings#Cut


```suggestion
		key, val, ok := strings.Cut(label, "=")
        key, val = strings.TrimSpace(key), strings.TrimSpace(val)
        if !ok || key == "" || val == "" {
            l.Warningf("Invalid additional label format: %s", label)
            continue
        }
```

manugarg (commented):
This PR does more than fix (e.g. it now allows x=a=b) Can you please update the description.

immanuwell:
@manugarg done, addressed both review follow-ups on the latest 5ebbe78e: updated the PR description to call out the first = behavior explicitly, and switched the parsing code to strings.Cut.
Re-ran go test ./surfacers/options and go test ./surfacers.

manugarg:
Thank you!

PR #1356, opened 2026-06-02
fix(config): check scanner.Err() after handleIncludes loop

### LINKED ISSUE #1355: Lines longer than 64KiB in a Cloudprober configuration file are silently ignored
**Describe the bug**

Lines in Cloudprober configuration files over 64KiB in length cannot be parsed by a `bufio.Scanner` are silently ignored

**Cloudprober Version**

0.14.2

**To Reproduce**

Take an existing configuration file for Cloudprober that is greater in size than 64KiB, and flatten it so it's all a single line.

Start Cloudprober using this file as configuration. No probes will be run, and querying the `/config-parsed` endpoint will show minimal/no config.

There is no reported error in the Cloudprober logs giving any indication that configuration was ignored or skipped.

**Additional context**

We machine-generate our cloudprober configuration files, and up until now have been writing these out via the [`.String()` method on a `ProberConfig`](https://github.com/cloudprober/cloudprober/blob/main/config/proto/config.pb.go#L170). This outputs all the configuration on a single line. 

Recentlly, when getting to about 55-60 probes in an individual configuration file, we have run into errors where Cloudprober has been executing zero probes on the workers where there are a larger number (55-60) of probes. Looking into this futher, it seems that `handleIncludes` makes use of a `bufio.Scanner` but doesn't explicitly check for error cases:

```go
func handleIncludes(baseDir string, content []byte) (string, error) {
	// snip...
	
	scanner := bufio.NewScanner(bytes.NewReader(content))
	for scanner.Scan() {
		line := scanner.Text()
		// handle line
		// snip...
	}

	// scanner.Err() is not called here
	// snip...
}
```

In this case, we are running into the `MaxScanTokenSize` for a `bufio.Scanner`: https://pkg.go.dev/bufio#pkg-constants. The `scanner` stores a `ErrTooLong` internally, but this needs to be explicitly checked for by calling `scanner.Err()`.

There are other error conditions that this code is susceptible to, not just our `ErrTooLong` case, and since these are never detected by calling `scanner.Err()`, they are not surfaced. Since `scanner.Scan()` also returns no output in the case where an error is generated, this results in no probes being parsed from the configuration file.

toller892:
## Description

`handleIncludes` in `config/config.go` uses a `bufio.Scanner` to iterate over config file lines, but never checks `scanner.Err()` after the scan loop. When a line exceeds `bufio.MaxScanTokenSize` (64KiB), the scanner silently stops — `Scan()` returns false and the remaining content is discarded with no error surfaced.

This affects machine-generated config files (e.g., output from `.String()` on `ProberConfig`) where all configuration is on a single line. Once the config grows past ~60 probes, the line exceeds 64KiB, the scanner stops, and Cloudprober starts with zero probes — with no error in the logs.

## Fix

Add `scanner.Err()` check after the scan loop:

```go
if err := scanner.Err(); err != nil {
    return "", fmt.Errorf("error reading config content: %w", err)
}
```

This surfaces the error (typically `bufio.ErrTooLong`) instead of silently producing an incomplete config.

Fixes #1355

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -127,6 +127,10 @@ func handleIncludes(baseDir string, content []byte) (string, error) {
 		}
 	}
 
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("error reading config content: %w", err)
+	}
+
 	newline := "\n"
 	if runtime.GOOS == "windows" {
 		newline = "\r\n"


manugarg:
I believe #1357 is better fix as it contains tests too.

PR #1358, opened 2026-06-08
[probes.http] Add opt-in OpenTelemetry tracing

cbroglie:
Cloudprober is the client in HTTP probes, so it now starts the root span for each sampled request and exports it to an OTLP collector. This makes probe requests visible in distributed traces and propagates W3C trace context to the target.

Tracing is configured globally via a new top-level `tracing` block and is off by default. When configured, all tracing-capable probes (HTTP today) are instrumented automatically; there are no per-probe tracing options.


cbroglie:
Happy to make any tweaks to the config/tracer setup, I just made some arbitrary choices. But the functionality is working as expected in my testing.

manugarg:
@cbroglie I was out on vacation for the last few weeks. I'll take a look at it in the next few days.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,187 @@
...
+		sdktrace.WithResource(res),
+	)
+
+	// Build the propagator from the configured names. When none are configured,
+	// fall back to OTEL_PROPAGATORS, defaulting to tracecontext and baggage.
+	prop := autoprop.NewTextMapPropagator()
+	if names := config.GetPropagator(); len(names) > 0 {
+		prop, err = autoprop.TextMapPropagator(names...)
+		if err != nil {
+			return nil, fmt.Errorf("invalid propagator config: %v", err)
```
manugarg, line 159: **[bug]** After `getExporter` succeeds and `NewTracerProvider` is created (which starts the batch span processor against `exp`), failure paths before returning `tp.Shutdown` leak those resources. The invalid-propagator path at line 178 returns without shutting down `tp`/`exp`; the `resource.New` failure path at line 160 likewise abandons the exporter. `TestInitInvalidPropagator` exercises this leak on every run.

**Suggestion:** Validate propagator names before constructing the provider, or on any error after `getExporter`/`NewTracerProvider` call `tp.Shutdown` (or at least `exp.Shutdown`) before returning. Same pattern for the resource creation error path.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,187 @@
...
+		expConf := config.GetOtlpHttpExporter()
+
+		var opts []otlptracehttp.Option
+
+		if expConf.GetEndpointUrl() != "" {
+			u, err := url.Parse(expConf.GetEndpointUrl())
+			if err != nil {
+				return nil, fmt.Errorf("invalid http endpoint_url: %s, err: %v", expConf.GetEndpointUrl(), err)
+			}
+			opts = append(opts, otlptracehttp.WithEndpoint(net.JoinHostPort(u.Hostname(), u.Port())))
```
manugarg: **[bug]** For HTTP OTLP exporters, a URL without an explicit port (e.g. `http://otel-collector` or `https://otel.example.com/v1/traces`) is turned into `net.JoinHostPort(host, "")` → `"host:"`, which is not a valid `WithEndpoint` value (`host:port`). Export will fail at runtime. Default OTLP/HTTP port is 4318 when omitted; this also differs from passing the full URL via env vars, where the SDK applies defaults correctly.

**Suggestion:** If `u.Port()` is empty, default to `4318` (or omit `WithEndpoint` and prefer `WithEndpointURL` if available on the pinned SDK, which accepts a full URL and applies scheme/path/port defaults). Add a unit test for endpoint URLs with and without ports.

probes/http/http.go
```diff
@@ -191,6 +207,8 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 			totalDuration, p.opts.Interval)
 	}
 
+	p.enableTracing = p.opts.ProberConfig.GetTracing() != nil && !tracing.SDKDisabled()
```
manugarg: `enableTracing` is true whenever a top-level `tracing` config exists (and SDK is not disabled), independent of `sampling_fraction`. With the default `sampling_fraction = 0`, spans are not recorded/exported, but `otelhttp` still wraps every HTTP probe transport and injects `traceparent`/`baggage` (flags unsampled) on every request. Should we treat `sampling_fraction <= 0` as fully disabled for probe wrapping (and skip provider setup).

internal/tracing/tracing.go
```diff
@@ -0,0 +1,187 @@
...
+		if expConf.GetHttpHeader() != nil {
+			opts = append(opts, otlptracegrpc.WithHeaders(expConf.GetHttpHeader()))
+		}
+
+		if expConf.GetCompression() == configpb.Compression_GZIP {
+			opts = append(opts, otlptracegrpc.WithCompressor("gzip"))
+		}
+
+		if expConf.GetInsecure() {
+			opts = append(opts, otlptracegrpc.WithInsecure())
```
manugarg, line 112: **[suggestion]** gRPC exporter options allow both `insecure: true` and `tls_config` to be set. Both `WithInsecure()` and `WithTLSCredentials(...)` are appended; the resulting client credentials behavior is ambiguous/order-dependent and is an easy misconfiguration (especially when copying the example’s `insecure: true` into a TLS setup).

**Suggestion:** Reject conflicting configs (`insecure` + `tls_config`) with a clear Init error, or document precedence and enforce one path only.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,187 @@
...
+	exp, err := getExporter(ctx, config)
+	if err != nil {
+		return nil, err
+	}
+
+	attrKVs := []attribute.KeyValue{attribute.String("service.name", config.GetServiceName())}
+	for _, attr := range config.GetResourceAttribute() {
+		attrKVs = append(attrKVs, attribute.String(attr.GetKey(), attr.GetValue()))
+	}
+	res, err := resource.New(ctx, resource.WithHost(), resource.WithAttributes(attrKVs...), resource.WithFromEnv())
```
manugarg: **[nit]** Resource attributes are applied as `WithAttributes(service.name, …)` then `WithFromEnv()`. Depending on the SDK merge rules, `OTEL_SERVICE_NAME` / resource env vars can override the configured `service_name` and resource attributes, which may surprise operators who set both. Ordering is the reverse of the OTEL metrics surfacer (`WithFromEnv` then `WithAttributes`).

**Suggestion:** Align order with the metrics surfacer (env first, config last wins)?

internal/tracing/proto/config.proto
```diff
@@ -0,0 +1,84 @@
...
+    optional string key = 1;
+    optional string value = 2;
+  }
+  repeated Attribute resource_attribute = 3;
+
+  // Fraction of probe requests to trace, in the range [0.0, 1.0]. Cloudprober
+  // is the root of the trace, so this is the probability that a given request
+  // starts a sampled trace. Defaults to 0.0, i.e. tracing is off until you
+  // raise the sampling rate.
+  optional double sampling_fraction = 4 [default = 0.0];
```
manugarg: I think this may trip people. If someone adds a tracing_config, the default expectation will be that tracing is enabled at that point.

How about using 1.0 as default? In most common cases, probe traffic will not be a whole lot.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2025 The Cloudprober Authors.
```
manugarg: super-nit: s/2025/2026

manugarg (commented):
Sorry for the late review.

cbroglie:
No worries! Addressed the feedback.

cloudprober.go
```diff
@@ -236,6 +239,15 @@ func initWithConfigSource(configSrc config.ConfigSource) error {
 		}
 	}
 
+	var tracingShutdown tracing.ShutdownFunc
+	if cfg.GetTracing() != nil {
+		tl := logger.NewWithAttrs(slog.String("component", "tracing"))
```
manugarg, line 244: **[bug]** `tracing.Init` installs a global TracerProvider (with a batch span processor and exporter) immediately, but `tracingShutdown` is only stored on the success path at the end of `initWithConfigSource`. If anything after tracing init fails (`initDefaultServer`, gRPC listener, `prober.Init`, etc.), the error path cancels `initCtx` and closes the default listener but never calls the tracing shutdown func. That leaks the batcher goroutine / exporter resources and leaves a live global TracerProvider set. A subsequent Init retry (allowed because `cloudProber.prober` is still nil) would install another provider without tearing down the first.

**Suggestion:** Capture shutdown as soon as `tracing.Init` succeeds, and invoke it on every early-return path after that point (or use a local `defer` that runs shutdown unless init fully succeeds and ownership is transferred to `cloudProber.tracingShutdown`). Mirror the existing partial-init cleanup for listeners.

probes/http/http.go
```diff
@@ -547,10 +568,43 @@ func (p *Probe) httpClient(target endpoint.Endpoint) *http.Client {
...
+			}),
+			otelhttp.WithSpanOptions(trace.WithAttributes(
+				attribute.String("probe", p.name),
+				attribute.String("target", target.Name),
+			)),
+			// Add connection-level span events (DNS, connect, TLS, etc.) via
+			// httptrace. WithoutHeaders ensures no request/response header
+			// values are recorded on spans, avoiding accidental capture of
+			// sensitive headers.
+			otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
```
manugarg, line 594: **[suggestion]** Enabling tracing always attaches `otelhttptrace.NewClientTrace`, while the probe already installs its own `httptrace.ClientTrace` in `doHTTPRequest` for `latency_breakdown` / keep-alive `connEvent` metrics. Go's `httptrace.WithClientTrace` composes hooks, so both should fire, but this interaction is untested. A future otelhttp/otelhttptrace change that replaces rather than composes context traces would silently break latency breakdown and conn-event counting whenever tracing is on.

**Suggestion:** Add a regression test that enables both tracing and `latency_breakdown` (and ideally keep-alive conn events) and asserts breakdown metrics still update. Optionally compose explicitly if you want to own the behavior.

probes/http/http.go
```diff
@@ -547,10 +568,43 @@ func (p *Probe) httpClient(target endpoint.Endpoint) *http.Client {
...
+			opts = append(opts, otelhttp.WithTracerProvider(p.tracerProvider))
+		}
+		if p.propagators != nil {
+			opts = append(opts, otelhttp.WithPropagators(p.propagators))
+		}
+		transport = otelhttp.NewTransport(transport, opts...)
 	}
-	return &http.Client{Transport: p.baseTransport, CheckRedirect: p.redirectFunc}
+
+	return &http.Client{Transport: transport, CheckRedirect: p.redirectFunc}
```
manugarg, line 607: **[suggestion]** `otelhttp.NewTransport` records HTTP client metrics via the global `MeterProvider` by default, not only traces. Today that usually hits the noop provider, so it is mostly extra per-request work, but if anything later installs a global meter provider (or an operator expects "tracing config" to mean traces only), probe traffic will suddenly emit client metrics without a separate opt-in.

**Suggestion:** Pass `otelhttp.WithMeterProvider(noop.MeterProvider{})` (or an explicit disabled meter) so this feature stays traces-only unless metrics are intentionally configured.

probes/http/http.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/cloudprober/cloudprober/common/oauth"
 	"github.com/cloudprober/cloudprober/common/tlsconfig"
 	"github.com/cloudprober/cloudprober/internal/httpreq"
+	"github.com/cloudprober/cloudprober/internal/tracing"
```
manugarg, line 39: **[suggestion]** `probes/http` imports `internal/tracing` solely for `SDKDisabled()` / config gating. That package pulls OTLP gRPC/HTTP exporters and `autoprop` (which itself pulls B3/Jaeger/AWS/OT propagators). Tracing is opt-in at runtime, but these deps are now linked into every binary that includes the HTTP probe, increasing size for the common "tracing off" path.

**Suggestion:** Move `SDKDisabled()` (or a tiny env helper) to a dep-light package, or inline the env check in the HTTP probe so exporter-heavy imports stay confined to `cloudprober` init / `internal/tracing` production setup.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Cloudprober is the root of the trace, so the root sampler decides
+	// whether a request is traced based on the configured fraction.
+	sampler := sdktrace.ParentBased(sdktrace.TraceIDRatioBased(config.GetSamplingFraction()))
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exp),
+		sdktrace.WithSampler(sampler),
+		sdktrace.WithResource(res),
+	)
```
manugarg, line 175: **[suggestion]** `sampling_fraction` is documented as being in `[0.0, 1.0]`, but `Init` never validates it. Values outside the range are clamped by `TraceIDRatioBased` (≤0 never, ≥1 always) with no config error, which can hide typos (e.g. `10` meaning 10% vs 100%). The HTTP probe additionally treats `sampling_fraction == 0` as "do not wrap transport," while `Init` still builds a full exporter/provider whenever the `tracing` block is present—slightly inconsistent and wasteful for a "configured but fully unsampled" setup.

**Suggestion:** Reject fractions outside `[0.0, 1.0]` at Init with a clear error. Consider skipping provider/exporter setup when the fraction is 0, or keep provider init but document the HTTP short-circuit.

internal/tracing/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+  optional Compression compression = 4;
+
+  // Whether to use insecure gRPC connection.
+  optional bool insecure = 5;
+}
+
+// TracingConfig enables and configures distributed tracing for Cloudprober.
+// When set, Cloudprober initializes a global OpenTelemetry TracerProvider that
+// exports spans to the configured OTLP collector. All probes that support
+// tracing are instrumented automatically and share the sampling rate below.
```
manugarg, line 50: **[nit]** `GRPCExporter.http_header` is a gRPC metadata map named like HTTP headers (copied from the metrics OTEL surfacer). Works, but is confusing in a new public config surface.

**Suggestion:** Prefer `header` / `metadata` for the gRPC exporter (with a comment that it maps to gRPC metadata), or document the name explicitly if keeping parity with the surfacer is intentional.

manugarg (commented):
Sorry @cbroglie for late reviews. Review agent came up with a few more things. Looks mostly right to me. Can you please take a look?

internal/tracing/proto/config.proto
cbroglie: I added a comment b/c I had followed this approach to remain consistent with the existing surfacer. But I can rename it still if that's preferred.

cloudprober.go
cbroglie: While fixing this I noticed the gRPC listener wasn't being closed, so my change includes a small refactor to handle that as well

cbroglie:
Sorry for the slow turnaround, but I think everything has been addressed now!

internal/tracing/tracing.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	if sf := config.GetSamplingFraction(); sf < 0 || sf > 1 {
+		return nil, fmt.Errorf("sampling_fraction must be in [0.0, 1.0], got %v", sf)
+	}
+
+	// Resource attributes are applied after WithFromEnv so that explicitly
+	// configured values (including service_name) take precedence over
+	// OTEL_RESOURCE_ATTRIBUTES/OTEL_SERVICE_NAME, matching the OTEL surfacer.
+	attrKVs := []attribute.KeyValue{attribute.String("service.name", config.GetServiceName())}
```
manugarg, line 146: **[bug]** `OTEL_SERVICE_NAME` can never take effect here.

`service_name` has a proto2 default of `"cloudprober"`, so `GetServiceName()` never returns empty, and `WithAttributes(...)` is applied last in `resource.New` (later options win on key conflict -- `Merge(a, b)`: "if there are common keys between a and b, then the value from b will overwrite the value from a"). So the env-derived `service.name` is always clobbered.

I confirmed it: with `OTEL_SERVICE_NAME=my-service`, the env-only resource is `[deployment.environment=prod host.name=... service.name=my-service]`, while `cfg.GetServiceName()` returns `"cloudprober"` with `cfg.ServiceName == nil`. The configured-looking value wins even though nothing was configured.

That contradicts the comment right above ("explicitly configured values (including service_name) take precedence over OTEL_RESOURCE_ATTRIBUTES/OTEL_SERVICE_NAME") -- there is no way to *not* explicitly configure it. It also diverges from the OTel metrics surfacer, which never sets `service.name` and therefore honors the env var.

**Suggestion:** append the attribute only when `config.ServiceName != nil`, and drop the proto default in favour of applying `"cloudprober"` only if the resulting resource has no `service.name`. A test with `OTEL_SERVICE_NAME` set would pin this down.

probes/http/http.go
```diff
@@ -547,10 +569,48 @@ func (p *Probe) httpClient(target endpoint.Endpoint) *http.Client {
...
+			otelhttp.WithSpanOptions(trace.WithAttributes(
+				attribute.String("probe", p.name),
+				attribute.String("target", target.Name),
+			)),
+			// Add connection-level span events (DNS, connect, TLS, etc.) via
+			// httptrace. WithoutHeaders ensures no request/response header
+			// values are recorded on spans, avoiding accidental capture of
+			// sensitive headers.
+			otelhttp.WithClientTrace(func(ctx context.Context) *httptrace.ClientTrace {
+				return otelhttptrace.NewClientTrace(ctx, otelhttptrace.WithoutHeaders())
```
manugarg, line 596: **[bug]** `WithoutHeaders()` keeps header values off spans, but otelhttp's own client attributes still record the full URL including the query string, so the same class of secret leaks through a different door.

otelhttp v0.71 (`internal/semconv/client.go`) strips only userinfo:

```go
userinfo := req.URL.User
req.URL.User = nil
u = req.URL.String()          // query string intact
req.URL.User = userinfo
```

My end-to-end run shows `url.full="https://example.com"` on every client span. HTTP probes routinely carry tokens and API keys in `relative_url` query params, so this deserves the same treatment the headers got -- either redact the query (an `otelhttp` filter / custom attribute handling, or strip `RawQuery` before the transport sees it) or call it out loudly in the config docs so operators know spans will carry their query strings.

cloudprober.go
```diff
@@ -156,6 +159,43 @@ func initDefaultServer(c *configpb.ProberConfig, l *logger.Logger) (net.Listener
...
+		}
+
+		s = grpc.NewServer(serverOpts...)
+		state.SetDefaultGRPCServer(s)
+	}
+	reflection.Register(s)
+	// register channelz service to the default grpc server port
+	service.RegisterChannelzServiceToServer(s)
+
+	grpcLn, err := net.Listen("tcp", net.JoinHostPort(serverHost, grpcPort))
```
manugarg, line 191: **[bug]** The extraction flipped the order of listen vs. server registration, and that's a small regression.

Before, `net.Listen` ran first and the server was only created and stored after it succeeded. Now the server is built, published globally via `state.SetDefaultGRPCServer(s)`, and `reflection.Register(s)` / `RegisterChannelzServiceToServer(s)` run *before* we know the port is available.

So if `net.Listen` fails (port busy), we return an error having left a fully registered server in global state. A subsequent `Init` retry -- still possible since `cloudProber.prober` is nil -- finds a non-nil `state.DefaultGRPCServer()`, skips creation, and calls `reflection.Register(s)` on it again. grpc's `Server.register` does `logger.Fatalf("grpc: Server.RegisterService found duplicate service registration for %q")`, which is a hard `os.Exit(1)`.

**Suggestion:** move the `net.Listen` back above the server creation/registration block so a listen failure leaves no global state behind. (The retry hazard also exists today for a failure *after* a successful listen, but no need to widen it.)

Thanks for spotting the missing `grpcLn.Close()` while you were in here, by the way -- good catch.

go.mod
```diff
@@ -23,21 +23,28 @@ require (
...
-	go.opentelemetry.io/otel/sdk/metric v1.44.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.71.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.71.0
+	go.opentelemetry.io/contrib/propagators/autoprop v0.71.0
+	go.opentelemetry.io/otel v1.46.0
+	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.46.0
+	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp v1.46.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.46.0
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.46.0
+	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric v1.46.0
```
manugarg, line 34: **[suggestion]** This PR bundles a fairly broad dependency bump that isn't required by the feature: otel 1.44 -> 1.46, contrib 0.61 -> 0.71, grpc 1.83.1 -> 1.83.2, protobuf 1.36.11 -> 1.36.12, x/net, x/crypto, x/mod, x/tools, testify 1.11.1 -> 1.12.1, and this line -- `stdoutmetric` v0.44.0 -> v1.46.0.

Two of those touch existing code paths rather than just the new one: `stdoutmetric` is what the existing OTel *metrics* surfacer uses, and the contrib jump moves `otelgrpc`/`otelhttp` under the existing gRPC and metrics paths. Everything passes, but bundled like this a tracing revert also reverts the upgrades, and a regression in either is hard to bisect.

**Suggestion:** split the upgrades into their own commit ahead of the feature commit (or a separate PR I can merge first), leaving this commit with just the genuinely new modules. For reference, binary size goes 89.6 MB -> 91.0 MB, which is fine.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,184 @@
...
+	"google.golang.org/grpc/credentials"
+)
+
+// ShutdownFunc flushes any buffered spans and shuts down tracing resources.
+type ShutdownFunc func(context.Context) error
+
+// SDKDisabled reports whether OpenTelemetry SDK functionality is disabled by
+// OTEL_SDK_DISABLED=true.
+//
+// Deprecated: use otelsdk.SDKDisabled directly; it doesn't pull in OTLP
```
manugarg, line 53: **[nit]** This is a brand-new function that ships already marked `Deprecated:`, and nothing calls it -- `probes/http` and the test both use `otelsdk.SDKDisabled` directly. Since the package is `internal/`, there's no compatibility reason to keep a shim. Just delete `SDKDisabled` here.

cloudprober.go
```diff
@@ -364,6 +398,14 @@ func Start(ctx context.Context) {
 		if grpcSrv != nil {
 			grpcSrv.Stop()
 		}
+		// Flush any buffered trace spans before exiting.
```
manugarg, line 401: **[nit]** Two small things about the shutdown ordering here.

The flush runs *before* `cloudProber.cancelInitCtx()` a few lines down, so probes can still be minting spans while the provider is shutting down; those spans are silently dropped. Cancelling first and flushing after would give the batcher a more complete final batch.

Also `cloudProber.tracingShutdown` is read here outside the lock that's taken below. That matches the existing `cancelInitCtx` read so it's not a new problem, just noting it while we're here.

cloudprober.go
```diff
@@ -288,6 +302,19 @@ func initWithConfigSource(configSrc config.ConfigSource) error {
 		logger.SetDefaultLogStore(ls)
 	}
 
+	var tracingShutdown tracing.ShutdownFunc
+	if cfg.GetTracing() != nil {
+		tl := logger.NewWithAttrs(slog.String("component", "tracing"))
+		tracingShutdown, err = tracing.Init(context.Background(), cfg.GetTracing(), tl)
```
manugarg, line 308: **[nit]** With tracing added, the failure unwind is now a hand-written three-way teardown (`tracingShutdown`, `grpcLn`, `ln`) repeated at two return sites, and getting the order/subset right is on whoever adds the next init step.

**Suggestion:** accumulate a `cleanup` closure (or a slice of them) as each resource is acquired, `defer` it, and clear it once ownership transfers to `cloudProber` at the end. Not blocking, but it'll age better.

internal/tracing/proto/config.proto
```diff
@@ -0,0 +1,85 @@
...
+import "github.com/cloudprober/cloudprober/common/tlsconfig/proto/config.proto";
+
+option go_package = "github.com/cloudprober/cloudprober/internal/tracing/proto";
+
+enum Compression {
+  NONE = 0;
+  GZIP = 1;
+}
+
+message HTTPExporter {
```
manugarg, line 14: **[nit]** `HTTPExporter`, `GRPCExporter` and `Compression` here are near-verbatim copies of `internal/surfacers/otel/proto/config.proto`. With two consumers now, it may be worth factoring the OTLP exporter config into a shared proto that both import.

Definitely a follow-up rather than this PR -- and the same goes for the `http_header` naming question below. Your consistency argument is fair; if we rename it we should rename both, and that's a separate change.

examples/tracing/cloudprober.cfg
```diff
@@ -0,0 +1,48 @@
...
+    http_probe {
+        scheme: HTTPS
+    }
+    interval_msec: 5000
+    timeout_msec: 2000
+}
+
+# Top-level tracing config. Its presence instruments all tracing-capable probes
+# (HTTP today). Without it, no spans are created. See:
+# https://cloudprober.org/
```
manugarg, line 22: **[nit]** Two things on the example.

The link is a bare `https://cloudprober.org/` placeholder -- there's no docs page for this new top-level config surface yet. I'll take care of the docs, so either drop the "See:" line or leave it and I'll fill it in.

More usefully: it'd be worth noting the span volume here. In my end-to-end run each probe request produced **7 spans** -- the `GET /` client span plus `http.getconn`, `http.dns`, `http.connect`, `http.tls`, `http.headers` and `http.send` from otelhttptrace. With `sampling_fraction` defaulting to 1.0 (my call, so this is on me), a 1s-interval probe over 100 targets is roughly 700 spans/sec. The 0.1 in this example already hints at it, but calling out the 7x multiplier from the connection-level sub-spans would set expectations. Worth considering whether those sub-spans should be opt-in separately from tracing itself.

manugarg (commented):
Thanks for sticking with this @cbroglie, and sorry for another round. I built the branch, ran the full test suite (all green), config-tested the new example, and ran it end-to-end against a fake OTLP/gRPC collector -- spans, resource attributes, and W3C propagation all work as advertised. The earlier feedback is genuinely addressed: propagator validated before the exporter is built, insecure+TLS rejected, `WithEndpointURL`, the `otelsdk` split, the noop meter provider, and the latency-breakdown composition test.

A few things still open; the first two are worth fixing before merge.

### Response body not closed on read error (can't inline, line isn't in the diff)

`probes/http/http.go:370`:

```go
respBody, err := io.ReadAll(resp.Body)
if err != nil {
    l.WithAttributes(p.dynamicHeaderAttrs(req)...).Warning(err.Error())
    return err          // body never closed
}
```

This is a pre-existing connection leak, but tracing turns it into a span leak too: otelhttp's `wrappedBody.Read` calls `span.End()` only on `io.EOF`; a non-EOF error just sets the status, and only `Close()` ends the span. So every read failure now also leaks an unended span that the batch processor never sees. A `defer resp.Body.Close()` right after the `client.Do` error check fixes both. Happy for this to be a separate PR if you'd rather keep this one focused.

### Verification notes

- `go test ./...` passes on the branch.
- `go run ./cmd/cloudprober -configtest -config_file examples/tracing/cloudprober.cfg` passes.
- End-to-end against a local OTLP/gRPC receiver: resource comes through as `env=review host.name=... service.name=cp-verify`, client spans are correctly parented roots with `probe`/`target` attributes, and `traceparent` reaches the target.
- Binary size: 89.6 MB -> 91.0 MB (+1.6%). Fine.

probes/http/http.go
cbroglie: I added a new config option to the HTTP probe definition to define which parameters are redacted so that they can be included by default (I didn't know what to name the field, happy to edit that).

The current otelhttp implementation makes it kind of annoying to do the redacting, but I saw https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9643 is already in-flight, so `redactingExporter` can be dropped when that lands.

cloudprober.go
cbroglie: I was thinking of implementing this pattern from the get-go, but wasn't sure of your preferences 👍

examples/tracing/cloudprober.cfg
cbroglie: I updated the docs but decided against adding another lever for the connection-level spans, since my assumption is anyone interested in tracing would be interested in those details as well. Happy to add it if you disagree.

probes/http/http.go
cbroglie: I'm also happy to align headers with URL params and use `WithRedactedHeaders` instead of `WithoutHeaders` so headers are included by default, if desirable

internal/tracing/proto/config.proto
manugarg: I'll tackle this in a follow up PR.

internal/tracing/proto/config.proto
manugarg: I'll follow up on consistency across.

cloudprober.go
```diff
@@ -365,13 +406,22 @@ func Start(ctx context.Context) {
 			grpcSrv.Stop()
 		}
 		cloudProber.cancelInitCtx()
+		// Flush any buffered trace spans before exiting.
+		if cloudProber.tracingShutdown != nil {
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			if err := cloudProber.tracingShutdown(shutdownCtx); err != nil {
+				slog.Warn("Error shutting down tracing", "err", err)
+			}
+			cancel()
```
manugarg: Let's convert this into a function:
```
// shutdownTracing flushes buffered spans, bounding how long we wait for the
// collector.
func shutdownTracing(shutdown tracing.ShutdownFunc) {
      if shutdown == nil {
              return
      }
      ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
      defer cancel()
      if err := shutdown(ctx); err != nil {
              slog.Warn("Error shutting down tracing", "err", err)
      }
}
```

then you can just call this function here:
```
shutdownTracing(cloudProber.tracingShutdown) 
```
 
and above at line 326. Currently line 326 calls shutdown without timeout.

probes/http/http.go
manugarg: @cbroglie Let's just drop traces redaction for now. Just add a comment to the redact_url_query_in_logs option that query params will go to traces regardless of this option if tracing is enabled.

manugarg (commented):
@cbroglie sorry for the back and forth here (sometimes AI review agents over-litigate). 

I'd like to request one more structural change, this time for simplification. Can you please drop the query-param redaction from this PR. It's complex and will not be used much (at least not right away). It may become part of Otel SDK soon and even if we have to add it ourselves, I'd probably implement it reusing the same code.

Concretely, to remove:
- internal/tracing/redact_exporter.go and redact_exporter_test.go
- internal/tracing/otelsdk/redact.go
- redact_query_params_in_traces (field 26) from probes/http/proto/config.proto, plus regen
- the spanAttrs marker block in probes/http/http.go
- the newRedactingExporter wrap in tracing.go

Thanks in advance!

cbroglie:
>I'd like to request one more structural change, this time for simplification. Can you please drop the query-param redaction from this PR. It's complex and will not be used much (at least not right away).

More than happy to drop that complexity 👍

PR #1359, opened 2026-06-18
[surfacers.postgres] use pgxpool in postgres surfacers to recreate connections after they …

markoposavec:
Hello!

I've had issues with the postgres surfacer where data would just stop being written to the database. After investigating the implementation I found that only a single pgx.Conn is created on startup. If that connection is closed by the server (e.g. postgres restart, network issues, or idle connection timeout), the surfacer stops writing permanently, the background goroutine keeps running and logs warnings, but every write attempt fails silently until the whole service is restarted.

Fixed by replacing pgx.Conn with pgxpool.Pool, which manages a pool of connections and automatically reconnects after a failure.


manugarg (approved):
Thank you! Good catch.

PR #1363, opened 2026-07-06
[probes] Add EXTERNAL_GRPC probe type: delegate probe runs to a gRPC sidecar

manugarg:
## Summary
- New `EXTERNAL_GRPC` probe type (`probes/grpcext`): per-target probe runs are delegated to a sidecar gRPC service over a small versioned contract; sidecar/infra failures are reported as `internal_errors`, separate from probe failures.
- New `sidecar` Go SDK: a probe type is one struct (`New`/`Probe`/`Close`) + `Serve`; SDK owns the gRPC server, health service, config decoding, per-target session cache (`state_handle`), and idle-TTL eviction.
- Runnable example (`examples/external_grpc`): sidecar with `http` and `tcp` probe types + demo config.

Prototype / draft — extension mechanism exploration (no forking to add probe types).

## Test plan
- `go test ./probes/grpcext/... ./sidecar/...` (new unit + e2e-over-UDS tests)
- Ran the example end-to-end: real probes through the sidecar, per-target sessions reused, sidecar kill shows up as `internal_errors` while `success`/`total` stay clean.


manugarg:
Pushed fixes from an xhigh code review of the prototype. The main semantic change: **sidecar/infra errors now count as probe failures** (total/success keep moving, so alerting works during a sidecar outage); `internal_errors` remains as the diagnostic counter separating "sidecar broken" from "target down", with reasons in logs. Also: RPC deadline (hung target) and SDK `New()` errors classify as target failures, validators now run over sidecar payload, `ValidateConfig` RPC fails fast at Init, RunOnce reuses sessions via a probe-level handle cache, and the SDK session cache got refcounting (no mid-probe eviction), interval-aware TTL, `InvalidateSession()`, panic-safe Close, and registration/socket-path validation.

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+	if sess := s.sessions[string(handle)]; len(handle) > 0 && sess != nil && sess.typeName == req.GetProbeType() {
+		sess.lastUsed = time.Now()
+		sess.inUse++
+		sess.minTTL = minTTL
+		s.mu.Unlock()
+		return sess, handle, nil
+	}
+	s.mu.Unlock()
+
+	val, err := h.newSession(ctx, t, req.GetConfig())
```
manugarg: **[bug]** `acquireSession` calls `h.newSession` (author `New`) without the panic recovery used for `Probe` (`runProbeSafely`) and `Close` (`closeSafely`). The package explicitly aims to keep a buggy probe type from taking down a multi-type sidecar, but a panic in `New` during first contact, stale-handle re-init, or one-shot setup is unprotected. gRPC-go does not reliably recover handler panics without an interceptor, so this can crash the whole sidecar process.

**Suggestion:** Wrap `newSession` the same way as `runProbeSafely` (recover → `Internal` / `newSessionErrorResult`), including the one-shot path at `Probe` that also calls `h.newSession` directly.

pkg/sidecar/sidecar.go
```diff
@@ -0,0 +1,243 @@
...
+
+// Fail reports a target-level failure: the probe ran and the target is not
+// healthy. This counts against the target's success ratio.
+func Fail(err error) *Result {
+	return &Result{err: err}
+}
+
+// Internal reports a sidecar/infra failure ("the probe broke", not "the
+// target is down"). Cloudprober records it separately so it doesn't pollute
+// the target's success ratio.
```
manugarg: **[bug]** `Internal`'s godoc says cloudprober records the failure "separately so it doesn't pollute the target's success ratio." That is incorrect relative to both `service.proto` and `probes/grpcext/grpcext.go`: every run (including internal/sidecar failures) increments `total` and does not increment `success`, so success ratio *is* affected (intentionally, so alerting on `total`/`success` still fires). Only the *additional* `internal_errors` metric separates causes. Authors reading the SDK comment may misuse `Internal` vs `Fail` expecting non-alerting behavior.

**Suggestion:** Align the comment with the proto/README/grpcext semantics: internal failures still count as probe failures for alerting; they *also* increment `internal_errors`.

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+		return fmt.Errorf("listening on %s: %v", s.addr, err)
+	}
+
+	stopSweep := make(chan struct{})
+	defer close(stopSweep)
+	go s.sweepIdleSessions(stopSweep)
+
+	grpcServer := grpc.NewServer()
+	pb.RegisterProberServer(grpcServer, s)
+	healthpb.RegisterHealthServer(grpcServer, health.NewServer())
```
manugarg: **[bug]** `Serve` registers the gRPC health service via `health.NewServer()` but never calls `SetServingStatus`. In grpc-go, an unset service returns `NotFound` from `Check`, so readiness/liveness probes against this sidecar will fail even while the process is healthy. The in-repo gRPC server (`internal/servers/grpc`) correctly sets SERVING for registered services after startup.

**Suggestion:** After registering services, set SERVING for `""` and/or `cloudprober.probes.grpcext.Prober` (and NOT_SERVING on shutdown if you add graceful stop).

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+
+	log.Printf("cloudprober sidecar: serving probe types %v on %s", s.typeNames(), s.addr)
+	return grpcServer.Serve(lis)
```
manugarg: **[bug]** When `grpcServer.Serve` returns (listener error / process exit path), `stopSweep` is closed but cached sessions are never closed. Author `Close` hooks (DB handles, connection pools, etc.) only run on idle eviction or explicit invalidation. Killing or restarting a long-lived sidecar therefore leaks per-target resources until the OS reaps the process—idle TTL does not help if Serve exits immediately.

**Suggestion:** On Serve exit (and ideally on a graceful `Stop`), lock the session map, delete all sessions, and `closeSafely` each one; consider accepting a `context.Context` or returning a stoppable server handle for signal integration.

probes/grpcext/grpcext.go
```diff
@@ -0,0 +1,303 @@
...
+	if p.c.GetServer() == "" {
+		return errors.New("external_grpc probe: server must be specified")
+	}
+	if p.c.GetProbeType() == "" {
+		return errors.New("external_grpc probe: probe_type must be specified")
+	}
+
+	// Sidecars are expected to be co-located (unix domain socket or
+	// localhost). TODO(manugarg): Add mTLS support for TCP servers.
+	conn, err := grpc.NewClient(p.c.GetServer(), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
manugarg, line 126: **[suggestion]** `Init` creates a `grpc.ClientConn` via `grpc.NewClient` but only stores the generated client; the conn is never closed when the probe's `Start` returns (context cancel / probe removed) or when `Init` fails after dial (e.g. sidecar rejects config). The first-party `probes/grpc` probe closes connections when `Start` ends. Over long-running processes that add/remove probes, this leaks client connections and related goroutines.

**Suggestion:** Store `*grpc.ClientConn` on `Probe`, close it at the end of `Start`, and close on `Init` error paths after the conn was created.

probes/grpcext/grpcext.go
```diff
@@ -0,0 +1,303 @@
...
+		return nil
+	}
+}
+
+// isInternalRPCError classifies a failed Probe RPC: DeadlineExceeded and
+// Canceled mean the probe ran out of time — almost always a hung target
+// (the sidecar handler shares our deadline), so we count it as a plain probe
+// failure. Anything else (Unavailable, Unimplemented, ...) means we couldn't
+// properly talk to the sidecar.
+func isInternalRPCError(err error) bool {
```
manugarg: **[suggestion]** `isInternalRPCError` treats all `DeadlineExceeded` / `Canceled` RPCs as non-internal ("hung target"). That is right when the sidecar accepted the RPC and shared the deadline, but wrong when the client never reached a healthy sidecar—e.g. TCP dial to a blackholed/firewalled address that only fails when the probe timeout fires. Those runs currently look like target failures (`internal_errors` stays 0), undermining the main observability goal of the feature for non-UDS deployments.

**Suggestion:** Prefer a short, separate connect/wait-for-ready budget, or treat pre-handler failures differently (e.g. `WaitForReady` + distinguish transport errors). At minimum document that TCP connect hangs are misclassified and recommend UDS / local health checks; add a test for "listener accepts but never handlers" vs "no route."

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+	idleTTL    time.Duration
+	probeTypes map[string]handler
+	initErrs   []error
+
+	mu       sync.Mutex
+	sessions map[string]*session // keyed by state handle
+}
+
+// Serve runs the sidecar server. It blocks until the listener fails.
+func Serve(opts ...Option) error {
```
manugarg, line 134: **[suggestion]** `Serve` blocks until the listener fails and provides no graceful shutdown, drain, or signal-friendly stop API. Combined with Issue 4, production sidecars cannot cleanly restart under k8s/systemd without abandoning sessions. Example binary is `log.Fatal(Serve(...))` only.

**Suggestion:** Add `ServeContext(ctx, ...)` or return `(Server, error)` with `GracefulStop` that marks health NOT_SERVING, stops the gRPC server, and closes sessions.

probes/grpcext/grpcext.go
```diff
@@ -0,0 +1,303 @@
...
+
+	if p.c.GetServer() == "" {
+		return errors.New("external_grpc probe: server must be specified")
+	}
+	if p.c.GetProbeType() == "" {
+		return errors.New("external_grpc probe: probe_type must be specified")
+	}
+
+	// Sidecars are expected to be co-located (unix domain socket or
+	// localhost). TODO(manugarg): Add mTLS support for TCP servers.
```
manugarg, line 125: **[suggestion]** Transport is always insecure, with only a TODO for mTLS. Config (which may contain credentials/query secrets) and session handles ride the same channel. Fine for default UDS co-location, but the config also advertises arbitrary gRPC targets (`localhost:9314`, remote hosts). A network-exposed sidecar is currently unauthenticated and plaintext.

**Suggestion:** Before non-prototype use: require UDS or loopback by default, or plumb TLS/`tlsconfig` like other cloudprober gRPC surfaces; document threat model explicitly in `config.proto`.

pkg/sidecar/sidecar.go
```diff
@@ -0,0 +1,243 @@
...
+	if len(bytes.TrimSpace(config)) == 0 {
+		return c, nil
+	}
+	if err := json.Unmarshal(config, &c); err != nil {
+		return c, configError{fmt.Errorf("decoding probe config: %v", err)}
+	}
+	return c, nil
+}
+
+func (h typedHandler[C, S]) validateConfig(config []byte) error {
```
manugarg, line 219: **[suggestion]** `ValidateConfig` only JSON-decodes into `C`. There is no author hook for semantic validation (required fields, port present, etc.). Misconfigs that unmarshal fine (e.g. example `tcp` with port 0) pass Init and then fail every cycle—example uses `Internal` for missing port, so they also spam `internal_errors`.

**Suggestion:** Optional `Validate func(C) error` on `ProbeType`, invoked from `validateConfig`; keep decode-only as the default.

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+	"fmt"
+	"log"
+	"net"
+	"os"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	pb "github.com/cloudprober/cloudprober/probes/grpcext/proto"
```
manugarg, line 32: **[suggestion]** The public sidecar SDK (`pkg/sidecar`) imports `probes/grpcext/proto`, coupling an extension-author dependency to an internal probe package path. That is awkward if the SDK is meant to be the stable surface for out-of-tree sidecars and for non-Go language stubs generated from the same contract.

**Suggestion:** Longer-term, host `service.proto` under a neutral path (e.g. `pkg/sidecar/proto` or `probes/grpcext/proto` re-exported from `pkg`) so the wire contract is the shared dependency, not the probe implementation package.

probes/grpcext/grpcext_test.go
```diff
@@ -0,0 +1,371 @@
...
+	assert.False(t, runReq.LastRun.Success)
+	assert.Error(t, runReq.LastRun.Error)
+
+	result := runReq.Result.(*probeResult)
+	assert.Equal(t, int64(1), result.total)
+	assert.Equal(t, int64(0), result.success)
+	assert.Equal(t, int64(1), result.internalErrors)
+}
+
+func TestProbeDeadline(t *testing.T) {
```
manugarg, line 247: **[suggestion]** Client/sidecar tests cover happy path, target failure, internal error, sidecar-down UDS, deadline, validators, state handle, one-shot, and ValidateConfig—but not: `New` panic isolation, Serve shutdown/`Close` of sessions, health SERVING, TCP connect-timeout classification, or conn close on probe stop. Those are exactly the lifecycle gaps above.

**Suggestion:** Add focused tests for panic-in-`New`, session `Close` on server stop, health check SERVING, and (if classification changes) blackhole connect vs in-handler deadline.

pkg/sidecar/server.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if r.err != nil {
+		resp.Error = r.err.Error()
+	}
+	return resp
+}
+
+func mintHandle() []byte {
+	b := make([]byte, 16)
+	rand.Read(b)
```
manugarg, line 484: **[nit]** `mintHandle` ignores the error from `crypto/rand.Read`. Failure is rare, but a zeroed handle would collide if it ever happened twice.

**Suggestion:** Check the error and fall back to a process-unique counter or return an internal error if minting fails.

examples/external_grpc/cloudprober.cfg
```diff
@@ -0,0 +1,34 @@
...
+#   go run ./sidecar
+probe {
+  name: "web_via_sidecar"
+  type: EXTERNAL_GRPC
+  targets {
+    host_names: "cloudprober.org,www.google.com"
+  }
+  interval_msec: 5000
+  timeout_msec: 4000
+
```
manugarg, line 12: **[nit]** Example config uses `unix:///tmp/...` while `Listen`/`config.proto` docs prefer `unix:/path` for Windows drive-letter compatibility. POSIX-only demo is fine, but the README does not mention the single-colon form that the SDK documents as the portable default.

**Suggestion:** Use `unix:/tmp/cloudprober-sidecar.sock` in the example (or note both forms and the Windows caveat in the example README).

pkg/sidecar/sidecar.go
manugarg: Fixed.

manugarg (commented):
Publishing comments.

pkg/sidecar/server.go
manugarg: Fixed.

pkg/sidecar/server.go
manugarg: Fixed

probes/grpcext/grpcext.go
manugarg: Fixed in 8c168053: `Probe` now stores `*grpc.ClientConn` on `Probe` and closes it when `Start` returns (after `UpdateTargetsAndStartProbes` unblocks on ctx cancellation, matching `probes/grpc`'s pattern) and on `Init` failure paths after a successful dial. Covered by `TestStartClosesConnOnExit` and `TestInitClosesConnOnRejectedConfig`, both verified against actual `conn.GetState()`.

pkg/sidecar/sidecar.go
manugarg: Fixed in 8c168053: added an optional `Validate func(c C) error` field on `ProbeType`, invoked from `validateConfig` after decode, so a semantic error fails the `ValidateConfig` RPC (and therefore probe `Init`) the same way a decode error does.

examples/external_grpc/cloudprober.cfg
manugarg: Fixed in 8c168053: example sidecar's default `-addr` flag and `cloudprober.cfg` now use `unix:/tmp/cloudprober-sidecar.sock`, matching the documented portable form.

probes/grpcext/grpcext_test.go
manugarg: Added in 8c168053: `TestServeNewPanicIsContained` (panic in New), `TestCloseAllSessions` + verified live that sessions Close on Serve exit, `TestServeHealthSERVING` (health check returns SERVING, verified live too), `TestStartClosesConnOnExit`/`TestInitClosesConnOnRejectedConfig` (client conn lifecycle). The one gap left open is the blackhole-connect-vs-in-handler-deadline test, which is tied to the TCP deadline classification comment below — deferred together with that fix.

pkg/sidecar/server.go
manugarg: Checked, not changing: this repo pins Go 1.25 (go.mod), and since Go 1.24 `crypto/rand.Read` is documented to never return an error (it crashes the program internally instead if the OS entropy source is unavailable), so the ignored return value can't actually cause a zeroed/colliding handle on this toolchain.

probes/grpcext/grpcext.go
manugarg: Agreed this is a real gap, not fixing in this pass: correctly classifying it needs a separate connect/wait-for-ready budget (or transport-error-vs-application-error distinction), which is a design decision rather than a quick fix. Leaving this open as a known limitation — documented in the PR discussion. Prototype is UDS-only by default, which sidesteps it for the common case.

pkg/sidecar/server.go
manugarg: Agreed, not fixing in this pass: a real graceful-shutdown/drain API (ServeContext, GracefulStop marking NOT_SERVING, session drain) is more than a quick fix and deserves its own design. Leaving open as a follow-up.

probes/grpcext/grpcext.go
manugarg: Agreed, not fixing in this pass: the TODO already flags this. Deliberately deferred — this is still a prototype, default/documented deployment is co-located UDS, and mTLS wiring is real scope (config surface + cert handling) that deserves its own change. Leaving open as a follow-up.

pkg/sidecar/server.go
manugarg: Fair point, not fixing in this pass: relocating service.proto to a neutral path is a real architectural improvement but a separate, broader change (touches the wire-contract's canonical location). Since nothing external depends on either path yet, there's no urgency forcing it now — leaving open as a follow-up.

probes/grpcext/grpcext.go
manugarg: Implemented in d972fbb2: `Probe.connLikelyDown()` checks `conn.GetState()` before sending — skips the RPC and classifies as `internal_error` if the state is exactly `TransientFailure` and this isn't the connection's first-ever request (deliberately not triggered by `Idle`, since long-interval probes legitimately go idle between cycles and reconnect transparently). Covered by `TestConnLikelyDownSkipsAmbiguousRPCs`. Went with the cheap state-check over the connect-budget-splitting approach I originally floated in this thread — full design discussion written up in the doc (grpc-probe-sidecar.md, "Classifying RPC failures" section) before implementing.

probes/grpcext/grpcext.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil {
+		if isInternalRPCError(err) {
+			result.internalErrors++
+		}
+		p.l.Warningf("target(%s): probe via sidecar %s failed: %v", target.Name, p.c.GetServer(), err)
+		runReq.LastRun.Set(false, 0, err)
+		return
+	}
+
+	if h := resp.GetStateHandle(); len(h) > 0 {
```
manugarg, line 264: **[suggestion]** Sidecar-side InvalidateSession correctly returns an empty `state_handle` so the client can drop the dead session (`pkg/sidecar/server.go` even documents that cloudprober "drops TargetState immediately"). The client only assigns `TargetState` when `len(h) > 0`, so an empty response leaves the previous handle in place. The next cycle still echoes the dead handle and forces the stale-handle re-New path. Functionally correct (session is recreated either way), but the claimed end-to-end fix for prior issue #11 is incomplete and the server comment is wrong about client behavior.

**Suggestion:** On a successful Probe RPC response, always apply the handle — including clearing when empty — e.g. set `runReq.TargetState = nil` when `len(resp.GetStateHandle()) == 0` (or always assign `resp.GetStateHandle()`). Add a client test that after an invalidate response with empty handle, the next request sends no `state_handle`.

probes/grpcext/grpcext.go
```diff
@@ -0,0 +1,360 @@
...
+// connection is always given the benefit of the doubt even if the state
+// isn't Ready yet — a cold connection is normal, not suspicious — but every
+// request after that isn't: a healthy connection that has worked before and
+// is now in TransientFailure means something changed.
+//
+// Deliberately not triggered by Idle: connections legitimately go idle
+// between cycles on longer-interval probes and reconnect transparently on
+// next use, so treating idle as "down" would misfire on exactly the
+// long-running probes this feature targets.
+func (p *Probe) connLikelyDown() bool {
```
manugarg: **[suggestion]** `connLikelyDown` correctly short-circuits only when `sentFirstRequest` is already true and `conn.GetState() == TransientFailure`, which is the unambiguous "gRPC already gave up connecting" window. That path is sound under concurrent multi-target use of a shared conn (all targets share one channel state). Residual gaps vs the original DeadlineExceeded misclassification remain by design: (1) the first probe request after Init never short-circuits even if the channel is already TransientFailure (and Init's ValidateConfig does not set `sentFirstRequest`, so "first request" means first runProbe, not first RPC on the conn); (2) Idle is deliberately ignored (good); (3) Connecting after a prior successful use (reconnect / blackhole TCP) can still burn the full probe timeout and land in `isInternalRPCError` as a non-internal DeadlineExceeded, i.e. a target failure. Metric shape on the short-circuit path (total++, success unchanged, internal_errors++) is correct.

**Suggestion:** Document these residual cases next to `connLikelyDown` / `isInternalRPCError` (or in the probe README) so operators know connect hangs mid-reconnect still look like target failures. Optional follow-up: treat DeadlineExceeded as internal when `GetState() != Ready` *after* the RPC returns (post-hoc check), which would catch Connecting hangs without false-positive on Idle-before-RPC. Not a prototype ship-blocker.

manugarg: Follow-up on the DeadlineExceeded classification fix (https://github.com/cloudprober/cloudprober/commit/d972fbb25be0f8b167c1ffacac946385a3f40876): grok suggested a post-hoc GetState() != Ready check as an optional refinement over the pre-send version. Verified gRPC's actual behavior with small repros before implementing anything further, since precision mattered here — and the pre-send version turned out to be checking the wrong condition.

A cleanly refused connection (closed port, bad UDS path) already fails fast with Unavailable via gRPC's default fail-fast semantics — measured ~18µs, not a wasted timeout. TransientFailure (the pre-send check's trigger) was never actually the ambiguous case. The real gap is a peer that accepts the TCP connection but never completes the gRPC handshake (hung sidecar process, misbehaving proxy): the channel sits in Connecting with no fast failure at all, so the RPC just blocks to our deadline and returns DeadlineExceeded — reproduced directly with a listener that accepts and never writes.

Replaced connLikelyDown() (pre-send, TransientFailure-triggered) with isInternalRPCError(err, connState) reading conn.GetState() after the RPC returns — reclassifies DeadlineExceeded/Canceled as internal if the transport wasn't Ready at that point. Simpler too: no "first request" bookkeeping needed, since it only judges after a real timeout already happened. Pushed in https://github.com/cloudprober/cloudprober/commit/237432093ac02316817e72ad19b1017a9c60391c, with a new test (TestInternalErrorOnHungHandshake) that reproduces the actual ambiguous scenario instead of a fast-failing one.

manugarg: Re-reviewed `23743209` — this addresses the residual gaps from the earlier `connLikelyDown` comment, and I'm satisfied.

The post-hoc rule (`DeadlineExceeded`/`Canceled` → internal iff `GetState() != Ready`) matches real gRPC behavior: a running handler implies a Ready transport, so a timeout while still Connecting/not-Ready is connect-time (sidecar/infra), not target. That correctly covers the hung-handshake case the pre-send TransientFailure short-circuit missed; fail-fast `Unavailable` on closed ports was never ambiguous.

Verified against `TestProbeDeadline` (handler hang → not internal) and `TestInternalErrorOnHungHandshake` (accept-never-write → internal). Remaining caveats (still burns full probe timeout on hung connect; tiny GetState race) are fine for a prototype and not misclassification bugs.

Resolving this thread.

probes/grpcext/grpcext.go
manugarg: Good catch — confirmed, this was a real gap. The client only assigned TargetState when the response handle was non-empty, so an explicit InvalidateSession clear never actually reached the client; it fell back to the (already-existing) unknown-handle recovery path instead, which is why it wasn't a correctness bug, just an incomplete fix.

Fixed in 96813c53: on a successful response, TargetState is now always set from the response handle, including clearing it to nil when empty. Added the test you asked for (TestProbeClearsStateHandleOnEmptyResponse) — after an invalidate response with an empty handle, the next request sends no state_handle.

PR #1364, opened 2026-07-06
[probes.sql] Add SQL probe type with POSTGRES flavor

manugarg:
## Summary
- New `SQL` probe type (`sql_probe`) with a flavor enum, POSTGRES only for now. Each run opens a fresh connection and runs the configured query (or just pings if no query), so latency covers connect + auth + query.
- Targets are optional (like browser probe): host/port come from the target when present, or the whole connection can be defined via `connection_string` (with `@target.xxx@` substitution). Query can also come from `query_file`.
- Query results are serialized one row per line and flow through the existing validators and payload-metrics machinery (`response_metrics_options`).

## Test plan
- Unit tests with a fake `database/sql` driver: `go test ./probes/sql/`
- Verified end-to-end against dockerized postgres 16: targets-based ping probe, 0-target query probe with GAUGE payload metrics + validator, wrong-password failure probe (success=0, clear error, no password leak), multiline `query_file` probe; `/config` and `/config-parsed` keep `envSecret` values masked.


probes/sql/postgres.go
```diff
@@ -0,0 +1,153 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("resolving target %s: %v", target.Name, err)
+			}
+			host = ip.String()
+		}
+
+		// Appending, rather than parsing-then-overwriting, lets pgx derive
+		// Fallbacks and TLS ServerName natively and correctly from the final
+		// host, instead of us patching its output after the fact.
+		inject := "host=" + host
```
manugarg: **[bug]** Target host (and, via the same path, resolved IP) is concatenated into the libpq key/value connection string without quoting/escaping: `inject := "host=" + host`. Libpq values are whitespace-delimited, so any space (or other special characters) in `target.Name` becomes additional connection parameters. For example a static target name of `db.example.com password=other sslmode=disable` yields `host=db.example.com password=other sslmode=disable`, overriding password/sslmode. Static `host_names` accept free-form names with no DNS-label validation, so this is reachable from config (and more generally from any targets source that can produce names with spaces). Port is safe because it is integer-formatted.

**Suggestion:** Always quote the injected host as a libpq value (single-quote and escape embedded quotes), e.g. `host='…'`, or stop string-appending and set `cfg.Host` / `cfg.Port` after `ParseConfig` while still fixing Fallbacks/TLS ServerName the way the comment intends. Add a unit test with a host containing spaces / `password=` text.

probes/sql/sql.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	if p.c.GetTlsConfig() != nil {
+		p.tlsConfig = &tls.Config{}
+		if err := tlsconfig.UpdateTLSConfig(p.tlsConfig, p.c.GetTlsConfig()); err != nil {
+			return fmt.Errorf("tls_config error: %v", err)
+		}
+	}
+
+	if p.c.GetResponseMetricsOptions() != nil {
```
manugarg, line 151: **[suggestion]** `Init` rejects validators on a ping-only probe (no query), but allows `response_metrics_options` without a query. A ping path returns a nil/empty body, so the payload parser runs every success and silently produces no useful metrics. That is inconsistent with the validator guard and easy to misconfigure.

**Suggestion:** Mirror the validator check: if `response_metrics_options` is set and `query`/`query_file` is empty, return an Init error such as `response_metrics_options require a query`.

probes/sql/sql.go
```diff
@@ -0,0 +1,309 @@
...
+			val := *(v.(*any))
+			if b, ok := val.([]byte); ok {
+				val = string(b)
+			}
+			fmt.Fprintf(&buf, "%v", val)
+		}
+		buf.WriteByte('\n')
+
+		if buf.Len() > maxQueryResultSize {
+			return nil, fmt.Errorf("query result exceeded %d bytes, consider adding a LIMIT to the query", maxQueryResultSize)
```
manugarg: **[suggestion]** The 1MB result cap is enforced only after a full row has been written into the buffer. A single oversized cell (or a row larger than 1MB) can still allocate well past the documented limit before the error is returned, so the cap is a soft bound rather than a hard memory ceiling.

**Suggestion:** Bound growth while formatting (e.g. write through a limited writer, or check `buf.Len()` after each column and fail early), or document that the cap is per completed row and a single cell may exceed 1MB before failure.

probes/sql/postgres.go
```diff
@@ -0,0 +1,153 @@
...
+// (filled or not) that we don't attempt to rewrite, and key/value form may
+// already specify one explicitly. Detection uses pgx's own parsing so that
+// quoted values (e.g. password='p host=x') don't false-positive: a sentinel
+// host prepended to the string survives only if the string doesn't set one
+// itself (libpq's last-value-wins rule). pgx doesn't map hostaddr to Host --
+// it ends up in RuntimeParams -- so look for it there. A parse error is
+// returned as such; connStr itself is what failed to parse.
+func connStringConflictsWithTarget(connStr string) (bool, error) {
+	s := strings.TrimSpace(connStr)
+	if strings.HasPrefix(s, "postgres://") || strings.HasPrefix(s, "postgresql://") {
```
manugarg: **[nit]** URI detection uses case-sensitive prefixes (`postgres://`, `postgresql://`). Unusual but valid schemes such as `Postgres://` or `POSTGRESQL://` skip the early conflict path and fall into key/value parsing, which can produce confusing parse/conflict behavior with real targets.

**Suggestion:** Compare with `strings.HasPrefix(strings.ToLower(s), "postgres://")` / `"postgresql://"` (or let pgx classify the string if there is a reliable API).

probes/sql/postgres.go
```diff
@@ -0,0 +1,153 @@
...
+		tlsCfg.ServerName = name
+	}
+}
+
+func (p *Probe) pgOpenDB(target endpoint.Endpoint) (*gosql.DB, error) {
+	cfg, err := p.pgConnConfig(target)
+	if err != nil {
+		return nil, err
+	}
+	return gosql.OpenDB(stdlib.GetConnector(*cfg)), nil
```
manugarg, line 170: **[nit]** Each run creates a fresh `database/sql.DB` via `OpenDB` without `SetMaxOpenConns(1)`. With a single Ping/Query this is usually one connection, but the pool defaults are unlimited; setting max open (and optionally max idle) to 1 would match the "one connect per probe run" model and avoid accidental multi-conn behavior if the code path grows.

**Suggestion:** After `OpenDB`, call `db.SetMaxOpenConns(1)` and `db.SetMaxIdleConns(0)` (or 1) before Ping/Query.

PR #1382, opened 2026-07-14
[probes.grpcext] Add TLS/mTLS support for sidecar connections

manugarg:
## Summary
- **Client**: `EXTERNAL_GRPC` probe config gets `tls_config` (reuses `common/tlsconfig.TLSConfig`, same surface as the grpc/http/sql probes). Unset ⇒ insecure, the co-located-socket default.
- **SDK**: new `sidecar.TLS(certFile, keyFile, clientCAFile)` option. A non-empty client CA requires a client cert (mTLS); empty serves plain server-side TLS. Author controls strictness via the CA arg (e.g. a flag).
- Runnable example gains `--tls_cert` / `--tls_key` / `--client_ca` flags.

Follow-up to #1363 — closes the plaintext-transport gap noted in that review.

## Test plan
- `go test -race ./probes/grpcext/... ./pkg/sidecar/...`
- New tests: real mTLS handshake end-to-end on both sides (client cert accepted; missing client cert rejected → internal error), plus cert/key load error paths.

### DIFF
--- a/examples/external_grpc/README.md
+++ b/examples/external_grpc/README.md
@@ -43,6 +43,23 @@ increment `internal_errors`, so you can tell a broken sidecar apart from a
 broken target; the failure reason is in cloudprober's logs. Try killing the
 sidecar to see it in action.
 
+## Serving over TLS / mTLS
+
+The unix-socket default is meant for a co-located sidecar. To reach one over
+the network, serve over TLS so probe configs (which may carry credentials)
+aren't sent in the clear. The example takes cert flags:
+
+```sh
+go run ./sidecar --addr=:9314 \
+  --tls_cert=server.crt --tls_key=server.key --client_ca=ca.crt
+```
+
+Passing `--client_ca` requires clients to present a cert signed by that CA
+(mutual TLS), so only cloudprober can send probes. On the cloudprober side,
+set `tls_config` in the probe (`ca_cert_file`, plus `tls_cert_file` /
+`tls_key_file` for the client cert). Omitting `--client_ca` serves plain
+server-side TLS.
+
 ## Writing your own sidecar
 
 A probe type is one struct — config in, result + metrics out; the SDK owns
--- a/examples/external_grpc/README.md
+++ b/examples/external_grpc/README.md
@@ -55,10 +55,36 @@ go run ./sidecar --addr=:9314 \
 ```
 
 Passing `--client_ca` requires clients to present a cert signed by that CA
-(mutual TLS), so only cloudprober can send probes. On the cloudprober side,
-set `tls_config` in the probe (`ca_cert_file`, plus `tls_cert_file` /
-`tls_key_file` for the client cert). Omitting `--client_ca` serves plain
-server-side TLS.
+(mutual TLS), so only cloudprober can send probes. Omitting `--client_ca`
+serves plain server-side TLS.
+
+On the cloudprober side, set `tls_config` in the probe:
+
+```
+probe {
+  name: "web_via_sidecar"
+  type: EXTERNAL_GRPC
+  targets { host_names: "cloudprober.org" }
+  external_grpc_probe {
+    server: "sidecar.example:9314"  # the sidecar, not the target
+    probe_type: "http"
+    tls_config {
+      ca_cert_file: "ca.crt"      # verifies the sidecar's server cert
+      tls_cert_file: "client.crt" # client cert (for mTLS)
+      tls_key_file: "client.key"
+      # server_name: "sidecar.example"
+    }
+  }
+}
+```
+
+`server_name` overrides the name checked against the sidecar's server cert.
+Set it when `server` is an IP or otherwise doesn't match a name/IP SAN on the
+cert (gRPC otherwise derives the expected name from the dial target, and
+verification fails) — e.g. `server: "10.0.0.5:9314"` with a cert issued for
+`sidecar.example` needs `server_name: "sidecar.example"`. Avoid
+`disable_cert_validation`: it skips verification entirely, which defeats the
+point of protecting credential-bearing probe configs on the wire.
 
 ## Writing your own sidecar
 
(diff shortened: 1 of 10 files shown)


examples/external_grpc/README.md
```diff
@@ -43,6 +43,23 @@ increment `internal_errors`, so you can tell a broken sidecar apart from a
...
+aren't sent in the clear. The example takes cert flags:
+
+```sh
+go run ./sidecar --addr=:9314 \
+  --tls_cert=server.crt --tls_key=server.key --client_ca=ca.crt
+```
+
+Passing `--client_ca` requires clients to present a cert signed by that CA
+(mutual TLS), so only cloudprober can send probes. On the cloudprober side,
+set `tls_config` in the probe (`ca_cert_file`, plus `tls_cert_file` /
```
manugarg: **[suggestion]** The client-side setup docs only list `ca_cert_file` / `tls_cert_file` / `tls_key_file` and never mention `server_name`, nor show a concrete `tls_config { ... }` snippet. When `server` is an IP (common for k8s/service IPs) and the leaf cert is hostname-based, gRPC derives ServerName from the dial target and cert verification fails unless `server_name` is set. The tests set `ServerName: "localhost"` for exactly this reason when dialing `127.0.0.1`, but operators reading the README would not know to do the same.

**Suggestion:** Document `server_name` (required when the dial target host does not match a name/IP SAN on the server cert) and include a minimal probe config example, e.g. `tls_config { ca_cert_file; tls_cert_file; tls_key_file; server_name: "sidecar.example" }`. Optionally note that `disable_cert_validation` defeats the point of protecting credential-bearing probe configs over the network.

examples/external_grpc/sidecar/main.go
```diff
@@ -107,11 +107,19 @@ var tcpProbe = sidecar.ProbeType[tcpConfig, any]{
...
 	flag.Parse()
 
-	log.Fatal(sidecar.Serve(
+	opts := []sidecar.Option{
 		sidecar.Listen(*addr),
 		sidecar.Register("http", httpProbe),
 		sidecar.Register("tcp", tcpProbe),
-	))
+	}
+	if *tlsCert != "" {
```
manugarg, line 126: **[suggestion]** TLS is enabled only when `--tls_cert` is non-empty. Passing `--client_ca` (or `--tls_key`) without `--tls_cert` is silently ignored and the process still serves plaintext — easy to misconfigure when the intent was mTLS on a network listener (`--addr=:9314`).

**Suggestion:** If `--client_ca` or `--tls_key` is set while `--tls_cert` is empty, fail fast with a clear flag error before `Serve`. Same if `--tls_cert` is set and `--tls_key` is empty (today this fails later inside `LoadX509KeyPair` with a less obvious path error).

pkg/sidecar/sidecar_test.go
```diff
@@ -443,3 +524,65 @@ func TestServeGracefulShutdown(t *testing.T) {
 	// The session left behind must have been closed on shutdown.
 	assert.Greater(t, closedSessions.Load(), closedBefore, "session was not closed on shutdown")
 }
+
```
manugarg, line 470: **[suggestion]** Coverage is strong for mutual TLS success and missing-client-cert rejection (both sidecar SDK and grpcext probe), plus cert-load init errors. There is no test that server-only TLS (`TLS(cert, key, "")` / empty client CA) accepts a client that presents no client cert and still encrypts/authenticates the server. That is an explicitly supported mode in docs (`Omitting --client_ca serves plain server-side TLS`) and is easy to regress if `ClientAuth` is ever set unconditionally.

**Suggestion:** Add a small `TestServeServerTLS` (and optionally a probe-side counterpart) that serves with empty `clientCAFile`, dials with RootCAs + ServerName and no client cert, and asserts Probe succeeds.

pkg/sidecar/server.go
```diff
@@ -78,6 +81,48 @@ func IdleTTL(d time.Duration) Option {
...
+		creds, err := loadServerTLS(certFile, keyFile, clientCAFile)
+		if err != nil {
+			s.initErrs = append(s.initErrs, err)
+			return
+		}
+		s.creds = creds
+	}
+}
+
+func loadServerTLS(certFile, keyFile, clientCAFile string) (credentials.TransportCredentials, error) {
```
manugarg: **[nit]** `loadServerTLS` relies on `tls.LoadX509KeyPair` for empty/missing path validation. Callers that pass empty `keyFile` (or empty `certFile` via the API rather than the example flags) get a low-level open error rather than an intentional validation message. Minor API polish only; behavior is still fail-closed.

**Suggestion:** Early-return a clear error if `certFile == "" || keyFile == ""` before attempting to load, e.g. `TLS requires both certFile and keyFile`.

examples/external_grpc/README.md
manugarg: Done in 26575ef5 — documented `tls_config.server_name` (with the IP-vs-SAN rationale) and added a concrete `external_grpc_probe { tls_config { ... } }` snippet, plus a caveat against `disable_cert_validation`.

examples/external_grpc/sidecar/main.go
manugarg: Done in 26575ef5 — `-tls_key`/`-client_ca` without `-tls_cert` now fails fast instead of silently serving plaintext.

pkg/sidecar/sidecar_test.go
manugarg: Done in 26575ef5 — added `TestServeServerTLS`: empty client CA accepts a client presenting no cert over an encrypted, server-authenticated channel.

pkg/sidecar/server.go
manugarg: Done in 26575ef5 — `loadServerTLS` now returns "TLS requires both certFile and keyFile" up front; covered by `TestTLSOptionEmptyPaths`.

PR #1383, opened 2026-07-15
[probes.starlark] Add tls_configs: per-call TLS selection via tls= kwarg

manugarg:
## Summary

- Add `tls_configs` (name -> `TLSConfig` map) to the Starlark probe, selected per call with a new `tls=` kwarg on the http builtins: `http.get(url, tls="legacy")`. `tls_config` stays the probe-wide default for calls that don't pass one.
- Motivation: a probe-wide `tls_config` is actively wrong once one script hits several hosts — `ca_cert_file` *replaces* the system CA pool (an internal CA breaks any public host the same script touches), `server_name` is a per-host SNI override, and `disable_cert_validation` is rarely meant for every host.
- Omitting `tls=` always means `tls_config`, never "the only `tls_configs` entry" — unlike `oauth.token(name="")`, a single entry must still be named, so adding one can't silently retarget the rest of the script.

## Test plan

- `go test ./probes/starlark/` — new `TestHTTP_TLSConfigs` covers named selection, omitted-`tls=`-uses-default, default + named in one script, unknown name, and no-`tls_configs`-configured.


docs/content/docs/how-to/starlark-probe/index.md
```diff
@@ -116,8 +116,8 @@ All scripts get a small, fixed set of builtins. No filesystem, network beyond
...
-| `http.post(url, headers=None, body=None, json=None, max_redirects=N, keep_alive=False)` | HTTP POST. Pass `json=` for an auto-encoded JSON body (sets `Content-Type`), or `body=` for a raw string/bytes. `max_redirects` and `keep_alive` match `http.get`. |
+| `http.get(url, headers=None, max_redirects=N, keep_alive=False, tls="")` | HTTP GET, returns a `Response`. `max_redirects=0` disables following; `max_redirects=N` follows up to N (omit for Go's default of 10). `keep_alive` defaults to `False` to match the HTTP probe -- every call exercises DNS/TCP/TLS setup, which is what you usually want from a prober. Pass `keep_alive=True` to reuse a pooled connection across calls in a chained API flow. `tls` selects one of the probe's `tls_configs` for this call (see below). |
```
manugarg: **[suggestion]** The builtins table shows `tls=""` as if it were the default parameter value for `http.get`/`http.post`. That conflicts with the carefully implemented semantics where an explicit empty string is an error, and omitting `tls` means system-default TLS. Readers skimming the signature can easily conclude that `tls=""` is a no-op equivalent to omitting the kwarg — the exact footgun the empty/`None` guards exist to prevent. The later TLS section correctly documents the fail-loud behavior; the table does not.

**Suggestion:** Show the kwarg as optional without implying an empty-string default, e.g. `tls=...` / omit it from the default list, or document it as `tls=<name>` with a note that omitting means system defaults and that `""`/`None` error.

manugarg: Fixed in 763f57bd — the table now shows `tls=<name>` and notes inline that omitting means system-default TLS while `""`/`None` error.

PR #1386, opened 2026-07-23
[probes.starlark] Add dns builtin: dns.resolve -> DnsResult

manugarg:
## Summary
- Add a `dns` builtin: `dns.resolve(name, type="A", server=None, timeout=None) -> DnsResult` with `.answers` (flat `list[str]`), `.name`, `.type`, and a `.latency(unit="s")` method (`"s"`/`"ms"`). A result is truthy iff it found something.
- Thin wrapper over Go's `net.Resolver` (no `probes/dns`/`miekg` dep). `server=None` resolves the way the host does; an explicit `server` dials that nameserver with the pure-Go resolver. Finding nothing is a result (empty `.answers`); every other lookup failure raises.
- Deliberately a *resolver*, not a DNS client: no wire rcodes, no TTLs, NODATA indistinguishable from NXDOMAIN. Asserting real server behavior stays the `dns` probe type's job — docs state the split.
- `latencyMethod` helper lives in `builtins.go` so http Response `.latency()` can reuse the same call shape later.

## Test plan
- `go test ./probes/starlark/` — `starlark_dns_test.go` covers A/TXT records, lowercase type normalization, NXDOMAIN and NODATA as results, `server=None`, unsupported-type error, the `normalizeDNSServer` host/port/IPv6 matrix, timeout parsing, and the latency unit selector against an in-process DNS server.
- Docs: `dns.resolve` added to the builtins table plus a `dns` section in the Starlark probe how-to; the example there was run against the test server to verify it works.

### DIFF
--- a/probes/starlark/builtins.go
+++ b/probes/starlark/builtins.go
@@ -18,9 +18,47 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"time"
 
 	starlarklib "go.starlark.net/starlark"
 )
+
+// latencyMethod builds a Starlark method that reports a duration in a
+// caller-chosen unit: latency() -> float seconds (default), latency("ms") ->
+// milliseconds. Shared by result values that expose a measured latency
+// (dns.resolve today; http Response next) so the call shape is identical.
+func latencyMethod(recvName string, d time.Duration) *starlarklib.Builtin {
+	return starlarklib.NewBuiltin(recvName+".latency", func(_ *starlarklib.Thread, _ *starlarklib.Builtin, args starlarklib.Tuple, kwargs []starlarklib.Tuple) (starlarklib.Value, error) {
+		unit := "s"
+		if err := starlarklib.UnpackArgs(recvName+".latency", args, kwargs, "unit?", &unit); err != nil {
+			return nil, err
+		}
+		switch unit {
+		case "s":
+			return starlarklib.Float(d.Seconds()), nil
+		case "ms":
+			return starlarklib.Float(float64(d.Nanoseconds()) / 1e6), nil
+		}
+		return nil, fmt.Errorf("%s.latency: unknown unit %q (want \"s\" or \"ms\")", recvName, unit)
+	})
+}
+
+// optionalDurationSeconds converts a Value bound by UnpackArgs (with "??"
+// suffix) into a duration, interpreting the number as seconds. The bool result
+// is false when the kwarg was omitted or None.
+func optionalDurationSeconds(v starlarklib.Value, name string) (time.Duration, bool, error) {
+	if v == nil {
+		return 0, false, nil
+	}
+	switch n := v.(type) {
+	case starlarklib.Int:
+		i, _ := n.Int64()
+		return time.Duration(i) * time.Second, true, nil
+	case starlarklib.Float:
+		return time.Duration(float64(n) * float64(time.Second)), true, nil
+	}
+	return 0, false, fmt.Errorf("%s: expected a number of seconds, got %s", name, v.Type())
+}
 
 // sortedNames returns m's keys in sorted order. Used by the builtins that
 // select a probe-configured thing by name (oauth_configs, tls_configs) to list
@@ -40,6 +78,7 @@ func builtins(vars map[string]string) starlarklib.StringDict {
 		"log":          logModule(),
 		"state":        stateModule(),
 		"oauth":        oauthModule(),
+		"dns":          dnsModule(),
 		"print_metric": starlarklib.NewBuiltin("print_metric", printMetric),
 	}
 }
--- a/probes/starlark/builtins.go
+++ b/probes/starlark/builtins.go
@@ -45,19 +45,29 @@ func latencyMethod(recvName string, d time.Duration) *starlarklib.Builtin {
 
 // optionalDurationSeconds converts a Value bound by UnpackArgs (with "??"
 // suffix) into a duration, interpreting the number as seconds. The bool result
-// is false when the kwarg was omitted or None.
+// is false when the kwarg was omitted or None. Non-positive values are an
+// error: callers use this for timeouts, where 0 or negative would cancel the
+// context immediately with a confusing deadline error.
 func optionalDurationSeconds(v starlarklib.Value, name string) (time.Duration, bool, error) {
-	if v == nil {
-		return 0, false, nil
-	}
+	var d time.Duration
 	switch n := v.(type) {
+	case nil:
+		return 0, false, nil
 	case starlarklib.Int:
-		i, _ := n.Int64()
-		return time.Duration(i) * time.Second, true, nil
+		i, ok := n.Int64()
+		if !ok {
+			return 0, false, fmt.Errorf("%s: int does not fit in int64", name)
+		}
+		d = time.Duration(i) * time.Second
 	case starlarklib.Float:
-		return time.Duration(float64(n) * float64(time.Second)), true, nil
+		d = time.Duration(float64(n) * float64(time.Second))
+	default:
+		return 0, false, fmt.Errorf("%s: expected a number of seconds, got %s", name, v.Type())
 	}
-	return 0, false, fmt.Errorf("%s: expected a number of seconds, got %s", name, v.Type())
+	if d <= 0 {
+		return 0, false, fmt.Errorf("%s: expected a positive number of seconds, got %v", name, v)
+	}
+	return d, true, nil
 }
 
 // sortedNames returns m's keys in sorted order. Used by the builtins that
--- a/probes/starlark/builtins_dns.go
+++ b/probes/starlark/builtins_dns.go
@@ -1,0 +1,232 @@
+// Copyright 2026 The Cloudprober Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package starlark
+
+import (
+	"context"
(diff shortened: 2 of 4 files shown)


probes/starlark/builtins_dns.go
```diff
@@ -0,0 +1,232 @@
...
+
+func dnsResolve(thread *starlarklib.Thread, _ *starlarklib.Builtin, args starlarklib.Tuple, kwargs []starlarklib.Tuple) (starlarklib.Value, error) {
+	const fname = "dns.resolve"
+	var name, rtype string
+	var serverArg, timeoutArg starlarklib.Value
+	rtype = "A"
+	if err := starlarklib.UnpackArgs(fname, args, kwargs,
+		"name", &name,
+		"type?", &rtype,
+		"server?", &serverArg,
```
manugarg: **[bug]** `server` is declared with a single `?` optional marker (`"server?", &serverArg`), while `timeout` correctly uses `??`. With `?`, an explicit `server=None` is passed through as Starlark `None` into `optionalString`, which rejects non-string values (`expected string, got NoneType`). Omitting the kwarg works, but the PR body and the module comment both document `server=None` as “use system resolver config”. Callers writing `dns.resolve(name, server=None)` (or a variable that is `None`) will fail at runtime.

**Suggestion:** Change the UnpackArgs tag to `"server??", &serverArg` so omitted and `None` both leave a Go nil and `optionalString` treats them as “system default”, matching `timeout` and the documented API.

probes/starlark/builtins_dns.go
```diff
@@ -0,0 +1,232 @@
...
+
+	start := time.Now()
+	answers, err := dnsLookup(ctx, resolver, rtype, name)
+	latency := time.Since(start)
+
+	rcode := "NOERROR"
+	if err != nil {
+		var dnsErr *net.DNSError
+		if errors.As(err, &dnsErr) && dnsErr.IsNotFound {
+			rcode, answers = "NXDOMAIN", nil
```
manugarg: **[bug]** Any `*net.DNSError` with `IsNotFound` is reported as `rcode = "NXDOMAIN"`. Go sets `IsNotFound` for both true NXDOMAIN and NODATA / “name exists but no records of this type” (e.g. AAAA lookup on a name that only has A). Scripts that branch on `r.rcode == "NXDOMAIN"` will therefore misclassify NODATA. Relatedly, the package comment claims NXDOMAIN/NODATA responses become results with an accurate rcode story, but `net.Resolver` never exposes the wire rcode, so `"NOERROR"` / `"NXDOMAIN"` are a lossy reconstruction—not actual DNS rcodes (SERVFAIL/REFUSED never appear as `.rcode`; they raise).

**Suggestion:** Prefer a more honest signal: e.g. map all `IsNotFound` to a non-wire label like `"NOTFOUND"` (or expose `.not_found` / empty answers without claiming NXDOMAIN), or document clearly that `.rcode` is synthetic and only ever `"NOERROR"` or `"NXDOMAIN"` with NODATA collapsed into the latter. If true rcodes matter for the probe use case, this builtin cannot provide them via `net.Resolver` and would need a real DNS client.

probes/starlark/builtins_dns.go
```diff
@@ -0,0 +1,232 @@
...
+	"go.starlark.net/starlarkstruct"
+)
+
+// ----------------------------------------------------------------------------
+// dns module
+//
+// Thin wrapper around Go's net.Resolver (deliberately not a reuse of
+// probes/dns, mirroring how the http builtin stays independent of probes/http).
+// server=None uses the system resolver config; an explicit server routes the
+// pure-Go resolver at that address. A response with an NXDOMAIN/NODATA rcode is
```
manugarg: **[suggestion]** Comment says “only transport failures raise”. Implementation raises on any non-`IsNotFound` error from `dnsLookup` (timeouts, SERVFAIL, REFUSED, network errors, malformed responses, etc.). That is reasonable for a thin resolver wrapper, but the comment overstates the contract relative to the code.

**Suggestion:** Rephrase to something like “name-not-found (`IsNotFound`) is a result; all other lookup failures raise”, so script authors know SERVFAIL is an exception path, not `r.rcode`.

probes/starlark/builtins_dns.go
```diff
@@ -0,0 +1,232 @@
...
+		} else {
+			return nil, fmt.Errorf("%s: %v", fname, err)
+		}
+	}
+
+	return &dnsResult{name: name, rtype: rtype, answers: answers, rcode: rcode, latency: latency}, nil
+}
+
+// normalizeDNSServer appends the default DNS port when the server has none.
+func normalizeDNSServer(s string) string {
```
manugarg, line 141: **[suggestion]** `normalizeDNSServer` treats any string that fails `SplitHostPort` as a bare host and does `JoinHostPort(s, "53")`. That works for `"8.8.8.8"` and bare `"::1"`, but a bracketed IPv6 host without a port (`"[::1]"`) becomes `"[[::1]]:53"`, which is not dialable. Users who copy resolver addresses in bracket form will get opaque dial errors.

**Suggestion:** Strip surrounding `[]` before re-joining, or document that IPv6 servers must be either bare (`::1`) or include a port (`[::1]:53`). A small unit test matrix for host / host:port / IPv6 forms would lock this down.

probes/starlark/builtins.go
```diff
@@ -18,10 +18,48 @@ import (
...
+// optionalDurationSeconds converts a Value bound by UnpackArgs (with "??"
+// suffix) into a duration, interpreting the number as seconds. The bool result
+// is false when the kwarg was omitted or None.
+func optionalDurationSeconds(v starlarklib.Value, name string) (time.Duration, bool, error) {
+	if v == nil {
+		return 0, false, nil
+	}
+	switch n := v.(type) {
+	case starlarklib.Int:
+		i, _ := n.Int64()
```
manugarg: **[nit]** `optionalDurationSeconds` ignores the `ok` result of `Int.Int64()`. Values outside int64 range silently wrap when converted to `time.Duration`, producing a nonsense timeout instead of an error (unlikely in probe scripts, but inconsistent with `optionalInt` in `builtins_http.go`, which checks the fit).

**Suggestion:** Match `optionalInt`: if `!ok`, return a clear error (`timeout: int does not fit in int64`). Optionally also reject non-positive timeouts so `timeout=0` / negative do not immediately cancel the context with a hard-to-read deadline error.

probes/starlark/builtins_dns.go
manugarg: Fixed in d221fd4d — tag is now `"server??"`, so omitted and `None` both leave a Go nil. Added `TestDNS_ServerNoneUsesSystemResolver`.

probes/starlark/builtins_dns.go
manugarg: Documented rather than renamed (d221fd4d). `.rcode` is kept as-is but the package comment now states plainly that it is synthetic: `net.Resolver` never exposes a wire rcode, so it takes exactly two values (`"NOERROR"` / `"NXDOMAIN"`), NODATA is indistinguishable from true NXDOMAIN via `IsNotFound` and is reported as `"NXDOMAIN"`, and SERVFAIL/REFUSED raise instead of appearing as an rcode. Real rcodes would need a real DNS client; that stays out of scope for this builtin.

probes/starlark/builtins_dns.go
manugarg: Rephrased in d221fd4d: name-not-found is a result, every other lookup failure (timeout, SERVFAIL, REFUSED, network error, malformed response) raises. PR body updated to match.

probes/starlark/builtins_dns.go
manugarg: Fixed in d221fd4d — surrounding brackets are stripped before `JoinHostPort`, so `"[::1]"` and bare `"::1"` both normalize to `"[::1]:53"`. Added `TestNormalizeDNSServer` covering host / host:port / bare v6 / bracketed v6 / v6-with-port.

probes/starlark/builtins.go
manugarg: Both done in d221fd4d — `Int64()`'s `ok` is now checked (`int does not fit in int64`, matching `optionalInt`), and non-positive values are rejected so `timeout=0` fails with a clear message instead of an instantly-cancelled context. Covered by `TestOptionalDurationSeconds`.

manugarg:
Resolved the design tension in 30194a6d by removing `.rcode` rather than documenting around it.

The deciding argument: with `net.Resolver`, a lookup that finds nothing *always* comes back as `IsNotFound`, so `.rcode == "NXDOMAIN"` was exactly equivalent to `len(.answers) == 0`. The field carried no information and one false promise. Dropping it makes the surface match the implementation, which is what the review was really asking for.

- `DnsResult` is now `.answers` / `.name` / `.type` / `.latency()`. Truthiness becomes "found something" (previously `rcode == "NOERROR"`), so NODATA and NXDOMAIN are both falsey with empty answers — no longer a claim, just what happened.
- Backend stays `net.Resolver`. It isn't a compromise for the main use case: `server=None` resolves the way the host does (`/etc/hosts`, search list, ndots, cgo resolver), which is precisely "what does this machine see?" — and which a miekg-based client could not answer.
- No second builtin. The "assert real server behavior" capability already exists as the `dns` probe type; a second Starlark module would be that with a worse interface. Both the package comment and the docs now state the split explicitly.

Also addressed from the smaller notes:

1. Documented that `server=None` and `server=<addr>` are not the same lookup — the latter dials directly with the pure-Go resolver, skipping `/etc/hosts` and the host's resolver choice. It's a probe step, not `dig`.
2. Record-type surface left exactly as bounded; the `dnsTypes` comment already says SOA/CAA are out of scope without miekg.
3. Truthiness kept as the ergonomic branch, now off `.answers`.

Two things worth flagging beyond the review:

- **The builtin had no docs at all** — the original commit only touched Go files. Added a `dns.resolve` row to the builtins table and a `dns` section in `docs/content/docs/how-to/starlark-probe/index.md` covering the result shape, the two server modes, and the DNS-probe split.
- **The test DNS server wasn't answering authoritatively.** Go's pure resolver reads an empty non-authoritative reply as a lame referral and retries against the system resolvers, so the new NODATA test was silently leaving the sandbox and hitting real DNS. Fixed with `m.Authoritative = true`; the NODATA case now genuinely exercises the in-process server.

PR #1387, opened 2026-07-24
[probes/browser] Add internal_errors metric for browser-harness failures

manugarg:
## Summary
- Add an `internal_errors` metric on the browser probe: runs that fail because of the browser harness (missing / failed-to-launch / crashed browser, or the harness never starting) rather than the target — mirroring `EXTERNAL_GRPC`. It's still a failed run; `internal_errors` is an extra signal.
- Classification: an `onBegin` heartbeat marker (absent ⇒ harness never started) plus narrow launch/crash error signatures decided from final test outcomes; the probe reads marker files after the run and only counts on a non-zero exit. An `Init` preflight fails fast when `npx` or `@playwright/test` isn't installed.
- Not counted (documented limitation): non-launch worker crashes after the suite starts, and the pre-existing `outputDir`/`runID` reportDir-collision race under `requests_per_probe > 1`.

## Test plan
- `go test ./probes/browser/`

### DIFF
--- a/probes/browser/browser.go
+++ b/probes/browser/browser.go
@@ -52,6 +52,11 @@ import (
 
 const playwrightReportDir = "_playwright_report"
 
+// internalErrorMarkerFile is written into the playwright report dir by the
+// cloudprober reporter when a run fails because of the browser harness itself
+// (browser missing, browser failed to launch/crashed) rather than the target.
+const internalErrorMarkerFile = "cloudprober_internal_error"
+
 // Probe holds aggregate information about all probe runs, per-target.
 type Probe struct {
 	name string
@@ -93,6 +98,7 @@ var templates embed.FS
 type probeRunResult struct {
 	total             metrics.Int
 	success           metrics.Int
+	internalErrors    metrics.Int
 	latency           metrics.LatencyValue
 	validationFailure *metrics.Map[int64]
 }
@@ -118,6 +124,7 @@ func (prr probeRunResult) Metrics(ts time.Time, _ int64, opts *options.Options) 
 	em := metrics.NewEventMetrics(ts).
 		AddMetric("total", &prr.total).
 		AddMetric("success", &prr.success).
+		AddMetric("internal_errors", &prr.internalErrors).
 		AddMetric(opts.LatencyMetricName, prr.latency.Clone()).
 		AddLabel("ptype", "browser")
 
@@ -157,21 +164,23 @@ func (p *Probe) initTemplateFile(templates embed.FS, fileName string, data any) 
 func (p *Probe) initTemplates() error {
 	// Set up playwright config in workdir
 	data := struct {
-		TestDir            string
-		GlobalTimeoutMsec  int64
-		Screenshot         string
-		Trace              string
-		EnableStepMetrics  bool
-		DisableTestMetrics bool
-		Retries            int32
+		TestDir                 string
+		GlobalTimeoutMsec       int64
+		Screenshot              string
+		Trace                   string
+		EnableStepMetrics       bool
+		DisableTestMetrics      bool
+		Retries                 int32
+		InternalErrorMarkerFile string
 	}{
-		TestDir:            p.testDirPath(),
-		GlobalTimeoutMsec:  p.playwrightGlobalTimeoutMsec(),
-		Screenshot:         "only-on-failure",
-		Trace:              "off",
-		EnableStepMetrics:  p.c.GetTestMetricsOptions().GetEnableStepMetrics(),
-		DisableTestMetrics: p.c.GetTestMetricsOptions().GetDisableTestMetrics(),
-		Retries:            p.c.GetRetries(),
+		TestDir:                 p.testDirPath(),
+		GlobalTimeoutMsec:       p.playwrightGlobalTimeoutMsec(),
+		Screenshot:              "only-on-failure",
+		Trace:                   "off",
+		EnableStepMetrics:       p.c.GetTestMetricsOptions().GetEnableStepMetrics(),
+		DisableTestMetrics:      p.c.GetTestMetricsOptions().GetDisableTestMetrics(),
+		Retries:                 p.c.GetRetries(),
+		InternalErrorMarkerFile: internalErrorMarkerFile,
 	}
 	if p.c.GetSaveScreenshotsForSuccess() {
 		data.Screenshot = "on"
@@ -458,6 +467,13 @@ func (p *Probe) runPWTest(ctx context.Context, runReq *sched.RunProbeForTargetRe
 		}
 	}
 
+	// The cloudprober reporter writes internalErrorMarkerFile when the run
+	// failed because of the browser harness (browser missing/failed to launch)
+	// rather than the target. The process has fully exited by now, so the marker
+	// (if any) is on disk.
+	_, markerErr := os.Stat(filepath.Join(reportDir, internalErrorMarkerFile))
+	internalError := markerErr == nil
+
 	// We use startCtx here to make sure artifactsHandler keeps running (if
 	// required) even after this probe run.
 	p.artifactsHandler.Handle(p.startCtx, reportDir)
@@ -470,6 +486,9 @@ func (p *Probe) runPWTest(ctx context.Context, runReq *sched.RunProbeForTargetRe
 	}
 
 	runReq.LastRun.Set(err == nil, latency, err)
+	if internalError {
+		result.internalErrors.Inc()
+	}
 	if err != nil {
 		return
 	}
--- a/probes/browser/browser.go
+++ b/probes/browser/browser.go
@@ -23,6 +23,7 @@ import (
 	"embed"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -323,6 +324,10 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 		return fmt.Errorf("playwrightDir is not provided through config or PLAYWRIGHT_DIR env variable")
 	}
 
+	if err := preflightFn(p.c.GetNpxPath(), p.playwrightDir); err != nil {
+		return fmt.Errorf("browser harness preflight failed: %v", err)
+	}
+
 	p.workdir = p.c.GetWorkdir()
 	if p.c.GetWorkdir() == "" {
 		d, err := os.MkdirTemp("", "cloudprober_"+p.name)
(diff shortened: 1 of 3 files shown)


probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -1,12 +1,29 @@
...
 
+// Error substrings that indicate the browser harness itself failed (browser
+// missing, browser failed to launch, browser crashed) rather than the target
+// under test failing. These map to an internal_error rather than a probe
+// failure. In-test failures (e.g. a locator/navigation timeout) do not match
+// any of these and are treated as normal failures.
+const launchErrorSignatures = [
+  "browserType.launch",
+  "Executable doesn't exist",
+  "Browser has been closed",
```
manugarg: **[bug]** The signature `"Browser has been closed"` is far too broad. Playwright commonly reports `Target page, context or browser has been closed` for ordinary test/target failures (page closed after navigation, use-after-close, teardown races, site-induced tab close). That message contains this substring, so many **non-harness** failures will be misclassified as `internal_errors`, diluting the metric's value for "browser missing / failed to launch / crashed."

**Suggestion:** Drop `"Browser has been closed"` (or require it only together with a launch/crash prefix such as `browserType.launch` / `Target crashed`). Prefer narrower signatures that appear on real harness failures (e.g. `browserType.launch`, `Executable doesn't exist`, `Target crashed`, maybe `Browser closed unexpectedly` / `Protocol error` if validated against real logs). Add a few golden error-message fixtures for true vs false positives.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -24,10 +41,19 @@ const testLabels= (test: TestCase) => {
 }
 
 class CloudproberReporter implements Reporter {
+  // Set when the run fails outside the test body: a run-level error reported via
+  // onError, or a browser-launch failure attached to a test's before-hooks.
+  private internalError = false;
```
manugarg: **[bug]** `internalError` is sticky for the whole run. With retries enabled, a failed attempt that matches `isLaunchError` sets the flag; a later successful retry never clears it. `onEnd` then writes the marker even when `FullResult.status === "passed"` (flaky-then-pass). On the Go side, `cmd.Execute` returns nil, so the probe increments **both** `success` and `internal_errors` — which violates EXTERNAL_GRPC semantics (internal error implies a failed run: total moves, success does not).

**Suggestion:** In `onEnd`, only write the marker when `result.status !== "passed"` (and optionally only for `"failed"` / `"timedout"` / `"interrupted"`). Alternatively recompute from final outcomes instead of a sticky flag. Belt-and-suspenders on Go: only `internalErrors.Inc()` when `err != nil`.

probes/browser/browser.go
```diff
@@ -470,6 +486,9 @@ func (p *Probe) runPWTest(ctx context.Context, runReq *sched.RunProbeForTargetRe
 	}
 
 	runReq.LastRun.Set(err == nil, latency, err)
+	if internalError {
```
manugarg, line 534: **[bug]** Marker presence alone increments `internal_errors`, independent of process success. Combined with sticky reporter flag on retries (or any spurious marker write / reportDir collision), a successful Playwright exit can still bump `internal_errors` while also bumping `success`. EXTERNAL_GRPC never does that: `internal_errors` is only incremented on non-success paths.

**Suggestion:** Change to `if internalError && err != nil { result.internalErrors.Inc() }`. Optionally log when a marker exists but `err == nil` so sticky-reporter bugs surface.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -24,10 +41,19 @@ const testLabels= (test: TestCase) => {
...
+
   onBegin(config: FullConfig, suite: Suite) {
     info(`Starting the suite "${suite.title}" with ${suite.allTests().length} tests`);
   }
 
+  onError(error: TestError) {
+    warning(`Run-level error (not attached to a test): ${JSON.stringify(error)}`);
+    this.internalError = true;
```
manugarg: **[suggestion]** `onError` marks **every** global/worker error as an internal harness failure without inspecting the message. Playwright documents `onError` as covering "some global error, for example unhandled exception in the worker process," which includes test-author bugs and config mistakes, not only missing browsers / launch failures. That over-counts relative to the PR's stated goal ("browser missing, failed to launch, or crashed").

**Suggestion:** Apply the same (tightened) signature filter in `onError`, or maintain a separate allowlist for clearly infra messages (missing browser, failed to launch). If the intent is intentionally broader ("any out-of-test failure"), document that explicitly and accept that it is not a pure harness metric.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -1,12 +1,29 @@
...
+import * as path from 'path';
 
 const info = (string: string) => process.stderr.write("INFO "+string+'\n');
 const warning = (string: string) => process.stderr.write("WARNING "+string+'\n');
 
 const print = (string: string) => process.stdout.write(string+'\n');
 
+// Error substrings that indicate the browser harness itself failed (browser
+// missing, browser failed to launch, browser crashed) rather than the target
+// under test failing. These map to an internal_error rather than a probe
```
manugarg: **[suggestion]** Comment says launch/crash errors "map to an `internal_error` **rather than** a probe failure." That is incorrect relative to both this PR's Go code and EXTERNAL_GRPC: the run still fails (`total` increments, `success` does not); `internal_errors` is an additional signal.

**Suggestion:** Reword to match EXTERNAL_GRPC docs: still a failed run for alerting on total/success; `internal_errors` distinguishes harness/infra from target failure.

probes/browser/browser.go
```diff
@@ -458,6 +467,13 @@ func (p *Probe) runPWTest(ctx context.Context, runReq *sched.RunProbeForTargetRe
 		}
 	}
 
+	// The cloudprober reporter writes internalErrorMarkerFile when the run
+	// failed because of the browser harness (browser missing/failed to launch)
+	// rather than the target. The process has fully exited by now, so the marker
+	// (if any) is on disk.
+	_, markerErr := os.Stat(filepath.Join(reportDir, internalErrorMarkerFile))
```
manugarg: **[suggestion]** Design gap vs EXTERNAL_GRPC: failures that never get far enough to run the custom reporter never write a marker, so they are invisible to `internal_errors`. Examples: bad `npx` path, Playwright not installed, config load failure before reporters, process hard-killed by `CommandContext` cancel (SIGKILL) before `onEnd`. Those are classic harness/infra failures; EXTERNAL_GRPC counts transport/sidecar unavailability as `internal_errors`.

**Suggestion:** Document the limitation, and/or add a Go-side fallback when `err != nil` and stderr/exit text matches known harness patterns (or when the process never produced a report dir / never started Node). At minimum, treat "failed to start playwright process" distinctly if easy to detect from `exec` errors.

probes/browser/browser_test.go
```diff
@@ -557,3 +563,18 @@ func TestProbeComputeTestSpecArgs(t *testing.T) {
 		})
 	}
 }
+
+func TestProbeRunResultMetrics(t *testing.T) {
```
manugarg, line 655: **[suggestion]** Test coverage only checks that (1) the generated reporter source contains `onError` / the marker filename and (2) `Metrics()` exports the new field. There is no test that `runPWTest` increments `internalErrors` when the marker exists, no test that a normal failure without a marker does not, no test that success+marker interaction is correct, and no unit coverage of `isLaunchError` true/false cases (the actual classification logic).

**Suggestion:** Add a focused Go test that creates a report dir with/without `cloudprober_internal_error` and exercises the post-`Execute` accounting (mock/skip real Playwright). Extract or duplicate the signature list into a small table-driven check if you add a Node test, or at least document expected matching/non-matching error strings in a Go comment + golden fixtures reviewed in CI when the template changes.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -50,10 +76,30 @@ class CloudproberReporter implements Reporter {
...
+    if ((result.errors || []).some((e) => isLaunchError(e.message))) {
+      this.internalError = true;
+    }
     {{ if not .DisableTestMetrics }}
     print(`test_status{${testLabels(test)},status="${result.status}"} 1`);
     print(`test_latency{${testLabels(test)},status="${result.status}"} ${result.duration*1000}`);
     {{ end }}
   }
+
+  onEnd(result: FullResult) {
```
manugarg, line 117: **[suggestion]** `onEnd` ignores `FullResult.status`. Even without retries, writing the marker purely from a sticky flag is weaker than tying the signal to the run outcome Playwright already computed. Also, global timeout yields `status: "timedout"` — `onEnd` still runs (good), but hard kill from the probe timeout may not.

**Suggestion:** Gate marker write on non-passed full result status; consider `onExit` only if you need stronger flush guarantees (usually unnecessary if `onEnd` is awaited, which Playwright does).

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -50,10 +76,30 @@ class CloudproberReporter implements Reporter {
...
+    }
+    try {
+      fs.mkdirSync(reportDir, { recursive: true });
+      fs.writeFileSync(path.join(reportDir, "{{ .InternalErrorMarkerFile }}"), "1");
+    } catch (e) {
+      warning(`Failed to write internal error marker: ${e}`);
+    }
+  }
 }
 export default CloudproberReporter;
```
manugarg: **[nit]** File still ends without a trailing newline (`\ No newline at end of file` in the diff). Minor style/tooling annoyance.

**Suggestion:** Add a final newline.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed in 63bda91d: `onError` now sets `internalError` only when `isLaunchError(error.message)` matches. Other global/worker errors stay ordinary failed runs, and a harness that never started is caught by the missing heartbeat instead.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Reworded in 63bda91d to match EXTERNAL_GRPC semantics: "A matching run is still a failed run; the marker only tells the probe to also count it as an internal_error."

probes/browser/browser.go
manugarg: Addressed in 6d2c0d46 + 30b9694e. Added an `onBegin` heartbeat marker: a failed run with no heartbeat (bad npx, Playwright missing, config-load error, kill before `onBegin`) is now classified internal. Plus an Init preflight (`LookPath(npx)` + `@playwright/test` installed) that hard-fails at startup for the static, deploy-time cases. Browser-binary availability stays a per-run signal.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Added the trailing newline in 6d2c0d46.

probes/browser/browser.go
manugarg: Fixed in 6d2c0d46: `classifyInternalError` returns false when `err == nil`, so a successful run never counts `internal_errors` regardless of a stale/spurious marker or reportDir collision. Pinned by `TestClassifyInternalError`.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed two ways. Reporter: `onEnd` now returns early when `result.status === "passed"` (63bda91d), so a flaky-then-pass run is not marked. Go: `internal_errors` is only counted when `err != nil` (6d2c0d46), so `success` and `internal_errors` can no longer co-occur.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed in 63bda91d: `onEnd` returns early on `result.status === "passed"`, tying the marker to Playwright's computed outcome rather than only the sticky flag. The lost-onEnd-on-hard-kill case is covered separately by the Go-side heartbeat decision (6d2c0d46) plus listing the cloudprober reporter before html so its marker write wins the teardown race.

probes/browser/browser_test.go
manugarg: Added coverage: `TestClassifyInternalError` (table over marker/heartbeat/err combinations), `TestDefaultPreflight` (npx-missing / package-missing / ok), and extended reporter-source assertions (narrowed signatures present, `onError` filtered, `onEnd` gated, broad signature absent). Commits 6d2c0d46 / 30b9694e / 63bda91d.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed in 63bda91d: dropped `"Browser has been closed"`, keeping only `browserType.launch`, `Executable doesn't exist`, `Target crashed`, with a comment explaining the exclusion. A `NotContains` assertion guards against it creeping back.

manugarg:
Claude is really struggling with this PR. Closing it for now. I'll look into this later.

PR #1388, opened 2026-07-24
[probes.browser] Add internal_errors metric for missing-toolchain failures

manugarg:
## Summary
- Add an `internal_errors` counter to the browser probe (mirrors `external_grpc`): infra/environment failures still count as probe failures but are told apart from target-down.
- Preflight `npx` and the Playwright package in `Init`; missing → fatal error so the prober exits with a clear message (static install-time deps).
- Runtime: classify a missing browser binary (playwright's `Executable doesn't exist`) as `internal_errors`, via a new `ProcessStderr` callback in the command package (which now also drains streamed output before `Execute` returns).

This is PR 1 of 2. PR 2 will cover browser-launch-failure → global timeout, detected via the reporter.

## Test plan
- `go test ./probes/browser/ ./probes/common/command/` — new tests for preflight (npx/playwright missing), the browser-missing signature, and `ProcessStderr` plumbing.

### DIFF
--- a/probes/browser/browser.go
+++ b/probes/browser/browser.go
@@ -53,11 +53,14 @@ import (
 
 const playwrightReportDir = "_playwright_report"
 
-// browserMissingRe matches Playwright's error when the browser binary is not
-// installed, e.g. "browserType.launch: Executable doesn't exist at
-// /path/to/chromium... Please run the following command to download new
-// browsers: npx playwright install". We treat this as an internal error.
-var browserMissingRe = regexp.MustCompile(`Executable doesn't exist`)
+// internalErrorRe matches process-output signatures that indicate an
+// environment/infra problem rather than a target failure, so the run is
+// counted as an internal_error:
+//   - a missing browser binary -- Playwright's "browserType.launch: Executable
+//     doesn't exist at ..." error, and
+//   - an unresolvable Playwright package -- npx's "could not determine
+//     executable to run" error.
+var internalErrorRe = regexp.MustCompile(`Executable doesn't exist|could not determine executable to run`)
 
 // Probe holds aggregate information about all probe runs, per-target.
 type Probe struct {
@@ -319,14 +322,19 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 		return fmt.Errorf("playwrightDir is not provided through config or PLAYWRIGHT_DIR env variable")
 	}
 
-	// Preflight: npx and the Playwright package are static install-time
-	// dependencies -- if they're missing the probe can never run, so fail here
-	// (fatal) with a clear message rather than silently failing every run.
+	// Preflight: npx must be resolvable -- it's the command we exec, so a
+	// missing npx means the probe can never run. Fail here (fatal) with a clear
+	// message rather than silently failing every run.
 	if _, err := exec.LookPath(p.c.GetNpxPath()); err != nil {
 		return fmt.Errorf("npx not found (npx_path=%q): %v", p.c.GetNpxPath(), err)
 	}
+	// The Playwright package's location isn't guaranteed to be under
+	// playwrightDir: Node resolves it via NODE_PATH, the cwd walk-up, or a
+	// global install. So this is only a heads-up warning; if it's genuinely
+	// unresolvable at runtime, the run is counted as an internal_error (see
+	// internalErrorRe).
 	if _, err := os.Stat(filepath.Join(p.playwrightDir, "node_modules", "@playwright", "test")); err != nil {
-		return fmt.Errorf("playwright is not installed in %s (missing node_modules/@playwright/test): %v", p.playwrightDir, err)
+		p.l.Warningf("Playwright package not found under %s (node_modules/@playwright/test); relying on Node module resolution at runtime", p.playwrightDir)
 	}
 
 	p.workdir = p.c.GetWorkdir()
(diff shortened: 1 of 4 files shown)


probes/browser/browser.go
```diff
@@ -305,6 +322,21 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 		return fmt.Errorf("playwrightDir is not provided through config or PLAYWRIGHT_DIR env variable")
 	}
 
+	// Preflight: npx must be resolvable -- it's the command we exec, so a
+	// missing npx means the probe can never run. Fail here (fatal) with a clear
+	// message rather than silently failing every run.
+	if _, err := exec.LookPath(p.c.GetNpxPath()); err != nil {
```
manugarg, line 328: **[suggestion]** Missing `npx` is a fatal Init error. Because `prober.Init` aborts on the first `addProbe` failure, a single browser probe with a bad/missing `npx_path` prevents the entire cloudprober process from starting — not just that probe. That also means no continuous `total`/`success`/`internal_errors` series for the misconfigured env (unlike grpcext, which keeps running and surfaces infra failure via `internal_errors`). This is an intentional design choice for a static install-time dependency, but it is a sharp operational edge if other probes should keep running.

**Suggestion:** Keep the hard-fail if “fail fast with a clear message” is the product goal, but call it out in docs/changelog. Alternatively, match grpcext: warn in Init, let runs fail, and count them as `internal_errors` so other probes and metrics keep exporting.

probes/browser/browser.go
```diff
@@ -471,6 +513,9 @@ func (p *Probe) runPWTest(ctx context.Context, runReq *sched.RunProbeForTargetRe
 
 	runReq.LastRun.Set(err == nil, latency, err)
 	if err != nil {
+		if internalErr {
+			result.internalErrors.Inc()
```
manugarg, line 524: **[suggestion]** Tests cover the regex (`TestInternalErrorRe`), Init preflight (`TestInitPreflight`), and command-package stderr delivery/drain (`TestProcessStderr`) in isolation, but nothing asserts that a failed playwright run whose stderr matches `internalErrorRe` actually increments `result.internalErrors` while leaving `success` unchanged and still advancing `total`. A future regression in the `ProcessStderr` wiring, the `err != nil` guard, or the drain ordering could ship green.

**Suggestion:** Add a focused unit/integration test for the run path — e.g. drive `runPWTest`/`runProbe` with a stub command (or a test subprocess that writes `Executable doesn't exist at ...` to stderr and exits non-zero) and assert `total`/`success`/`internal_errors` and `LastRun` together.

probes/browser/browser.go
```diff
@@ -52,6 +53,15 @@ import (
...
 const playwrightReportDir = "_playwright_report"
 
+// internalErrorRe matches process-output signatures that indicate an
+// environment/infra problem rather than a target failure, so the run is
+// counted as an internal_error:
+//   - a missing browser binary -- Playwright's "browserType.launch: Executable
+//     doesn't exist at ..." error, and
+//   - an unresolvable Playwright package -- npx's "could not determine
+//     executable to run" error.
+var internalErrorRe = regexp.MustCompile(`Executable doesn't exist|could not determine executable to run`)
```
manugarg, line 63: **[suggestion]** The classifier is a loose substring match on `Executable doesn't exist` (and `could not determine executable to run`). Playwright’s real message is specifically `Executable doesn't exist at <path>`, but any test or app stderr containing the shorter phrase (assertions about missing binaries, custom CLIs, etc.) will mark a normal target failure as `internal_errors`. That can under-signal “target down” and over-signal “environment broken.”

**Suggestion:** Tighten to Playwright’s phrasing, e.g. `Executable doesn't exist at` (and keep `could not determine executable to run`). Optionally also match `browserType.launch:` for extra specificity.

probes/browser/browser.go
```diff
@@ -91,8 +101,14 @@ var templates embed.FS
 // (see documentation with statsKeeper below). That's the reason we use metrics.Int
 // types instead of metrics.AtomicInt.
 type probeRunResult struct {
-	total             metrics.Int
-	success           metrics.Int
+	total   metrics.Int
+	success metrics.Int
+	// internalErrors counts runs that failed because of environment/infra
```
manugarg, line 106: **[nit]** The `internalErrors` field comment says it counts “a missing browser binary today,” but `internalErrorRe` also classifies an unresolvable Playwright package (`could not determine executable to run`). Slightly stale relative to the regex and the Init warning path.

**Suggestion:** Update the comment to list both current signatures (missing browser binary and unresolved Playwright package / npx executable).

probes/common/command/command.go
```diff
@@ -104,6 +123,9 @@ func (c *Command) setupStreaming(cmd *exec.Cmd, l *logger.Logger) error {
 			} else {
 				l.WarningAttrs("process stderr", slog.String("process_stderr", scanner.Text()), slog.String("process_path", c.CmdLine[0]))
 			}
+			if c.ProcessStderr != nil {
+				c.ProcessStderr(scanner.Bytes())
```
manugarg, line 130: **[nit]** `ProcessStderr` is passed `scanner.Bytes()`, whose backing storage is reused on the next `Scan`. The browser callback only runs `regexp.Match` synchronously (safe), and `TestProcessStderr` copies via `string(line)` (safe). A future caller that retains the slice would race/corrupt data. The field docs don’t mention this.

**Suggestion:** Document that callbacks must not retain `line` without copying (same hazard as `ProcessStreamingOutput` / `bufio.Scanner`), or copy inside the package before invoking the callback.

probes/browser/browser.go
manugarg: This is the desired behavior.

probes/browser/browser.go
manugarg: This is acceptable risk.

probes/browser/browser.go
manugarg: Done in 02ca65bf — added `TestRunProbeInternalError`, which drives a full run through `runProbe` with a stub npx that writes Playwright's missing-browser error to stderr and exits non-zero, asserting `total`/`success`/`internal_errors` and `LastRun` together.

probes/browser/browser.go
manugarg: Done in 02ca65bf — comment now lists both signatures (missing browser binary and unresolvable Playwright package) and points to `internalErrorRe`.

probes/common/command/command.go
manugarg: Done in 02ca65bf — `ProcessStderr` doc now warns that the slice's backing array is reused on the next scan and must not be retained without copying.

PR #1389, opened 2026-07-24
[probes.browser] Count browser-launch timeouts as internal_errors

manugarg:
## Summary
- Detect the browser-launch-timeout case (browser can't launch under resource pressure → no progress → hits cloudprober's `globalTimeout`) and count it as `internal_errors` instead of an ordinary failure.
- Reporter `onEnd` emits a stderr sentinel `[cloudprober-internal-error]` when `result.status === "timedout"` **and** no test made real progress; browser.go's existing `internalErrorRe` classifier picks it up.
- `madeProgress` guard (a test passed, or failed with a populated error) prevents false positives — a genuine run that merely exhausted the budget stays a target/config failure.

Stacked on #1388 (PR-1); base is `browser-internal-errors-static`.

## Test plan
- `go test ./probes/browser/` — `TestInternalErrorRe` now covers the sentinel.
- Validated the rendered reporter against **real Playwright** (browserless tests): pass → no sentinel; forced `globalTimeout` with no progress → sentinel; pass-then-timeout → no sentinel (guard holds).

### DIFF
--- a/probes/browser/browser_test.go
+++ b/probes/browser/browser_test.go
@@ -277,7 +277,7 @@ func TestInternalErrorRe(t *testing.T) {
 	playwrightMissing := "npm error could not determine executable to run"
 	// Emitted by the cloudprober reporter's onEnd on a launch-timeout (see
 	// cloudprober-reporter.ts).
-	launchTimeout := "WARNING [cloudprober-internal-error] test suite timed out before the browser launched"
+	launchTimeout := "WARNING [cloudprober-internal-error] test suite timed out during browser launch"
 
 	assert.True(t, internalErrorRe.MatchString(browserMissing))
 	assert.True(t, internalErrorRe.MatchString(playwrightMissing))
@@ -323,7 +323,14 @@ func TestReporterDecision(t *testing.T) {
 	// and DisableTestMetrics are referenced by this template; the harness only
 	// sends pw:api steps and onEnd, so their values don't affect the outcome.
 	tmpl := template.Must(template.ParseFiles(filepath.Join("templates", "cloudprober-reporter.ts.tmpl")))
-	reporterPath := filepath.Join(t.TempDir(), "cloudprober-reporter.ts")
+	tmpDir := t.TempDir()
+	// node < 22.7 doesn't detect module syntax on its own, and picks the format
+	// from the nearest package.json; without this the stripped reporter's
+	// "export default" would be loaded as CommonJS and fail to parse.
+	if err := os.WriteFile(filepath.Join(tmpDir, "package.json"), []byte(`{"type":"module"}`), 0644); err != nil {
+		t.Fatal(err)
+	}
+	reporterPath := filepath.Join(tmpDir, "cloudprober-reporter.ts")
 	f, err := os.Create(reporterPath)
 	if err != nil {
 		t.Fatal(err)
--- a/probes/browser/browser_test.go
+++ b/probes/browser/browser_test.go
@@ -18,10 +18,13 @@ import (
 	"context"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
+	"text/template"
 	"time"
 
 	"github.com/cloudprober/cloudprober/metrics"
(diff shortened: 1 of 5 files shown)


probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -47,6 +55,13 @@ class CloudproberReporter implements Reporter {
   }
   
   onTestEnd(test: TestCase, result: TestResult) {
+    // A genuine execution is a pass, or a failure with a real error. A
+    // timeout-aborted test shows up as "failed" with an empty errors array
+    // (errors.length === 0), so it does not count as progress.
+    if (result.status === "passed" ||
```
manugarg: **[bug]** `madeProgress` is supposed to suppress the sentinel when a real run merely exhausts the budget, but under Playwright `globalTimeout` the in-flight test often never receives `onTestEnd` at all — so `madeProgress` stays `false` even if the browser launched and the test body was hanging on the target.

Empirical check (Playwright Test + custom reporter):
- hang mid-test, `globalTimeout` wins → `ON_TEST_BEGIN` only, no `ON_TEST_END`, suite `timedout`, `madeProgress=false` → sentinel **emitted**
- fail-then-hang → first `failed` sets progress → sentinel suppressed (works)

Cloudprober typically sets only `globalTimeout` (from probe timeout) and often runs a single smoke test. In that common path, a genuine target/page hang that hits the probe budget is miscounted as `internal_errors` — the opposite of the stated intent.

The comment that timeout-aborted tests show up as `"failed"` with empty `errors` is also inaccurate for this path.

**Suggestion:** Revisit the progress signal. Counting only terminal statuses in `onTestEnd` cannot distinguish launch hang vs mid-test hang when `onTestEnd` never runs. Stronger options:
1. treat any `onTestBegin` (or fixture/step evidence that a worker started the test) as progress, and only emit the sentinel when the suite `timedout` with **zero** `onTestBegin`s / zero tests started; or
2. detect launch-phase specifically (e.g. hang during `browserType.launch` / no browser context created) rather than “no completed test outcome”.

Document the intended false-positive/false-negative tradeoff once the heuristic is fixed.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -47,6 +55,13 @@ class CloudproberReporter implements Reporter {
...
+        (result.status === "failed" && result.errors.length > 0)) {
```
manugarg: **[bug]** Per-test timeouts finish as `result.status === "timedOut"` with a non-empty errors array (`Test timeout of Xms exceeded.`). That is real execution, but it is not counted as progress.

Empirical multi-test case:
1. first test `timedOut`
2. second test still running when suite hits `globalTimeout` (no `onTestEnd` for #2)
3. suite `timedout`, `madeProgress` still false → sentinel **emitted**

So an earlier completed timeout + later incomplete test is labeled a launch/internal failure.

**Suggestion:** At minimum count `timedOut` (and likely any completed non-skipped outcome) as progress:

```ts
if (result.status === "passed" ||
    result.status === "failed" ||
    result.status === "timedOut") {
  this.madeProgress = true;
}
```

Note this alone does **not** fix Issue 1 (single in-flight test with no `onTestEnd`).

probes/browser/browser_test.go
```diff
@@ -272,9 +272,13 @@ func TestInternalErrorRe(t *testing.T) {
 	browserMissing := "browserType.launch: Executable doesn't exist at /root/.cache/ms-playwright/chromium-1091/chrome-linux/chrome\n" +
 		"Please run the following command to download new browsers:\nnpx playwright install"
 	playwrightMissing := "npm error could not determine executable to run"
+	// Emitted by the cloudprober reporter's onEnd on a launch-timeout (see
+	// cloudprober-reporter.ts).
+	launchTimeout := "WARNING [cloudprober-internal-error] test suite timed out before the browser made progress"
 
 	assert.True(t, internalErrorRe.MatchString(browserMissing))
```
manugarg, line 282: **[suggestion]** Coverage only asserts that the sentinel string matches `internalErrorRe`. There is no automated test of the reporter `madeProgress` / `onEnd` decision table (timedout+no begin, timedout+passed earlier, timedout+failed earlier, timedout+timedOut earlier, pure per-test timeout → suite failed).

The false-positive risk above is exactly what a regex unit test cannot catch; the TS heuristic is the novel logic in this PR. Manual Playwright validation listed in the PR body covers pass / no-progress timeout / pass-then-timeout, but not "hanging after start until globalTimeout" (the common single-test probe case).

**Suggestion:** Extract a pure progress/sentinel decision helper and table-test it, and/or keep a small node script in CI that drives the reporter with synthetic events / real Playwright browserless hangs.

probes/browser/templates/cloudprober-reporter.ts.tmpl
```diff
@@ -47,6 +55,13 @@ class CloudproberReporter implements Reporter {
   }
   
   onTestEnd(test: TestCase, result: TestResult) {
+    // A genuine execution is a pass, or a failure with a real error. A
+    // timeout-aborted test shows up as "failed" with an empty errors array
```
manugarg: **[nit]** Comment claims a timeout-aborted test shows up as `"failed"` with `errors.length === 0`. Playwright’s statuses include `timedOut` and `interrupted`; under `globalTimeout` the in-flight test often never reaches `onTestEnd` at all. Misleading comments will make the next edit to this heuristic worse.

**Suggestion:** Rewrite to name real statuses / lifecycle and the product distinction (infra hang vs target/config slowness).

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed in 6a8bed9d by replacing `madeProgress` with `browserReady`: set when any `pw:api` step completes without error (the first such step is the browser launch), so onEnd emits the sentinel only when the suite timed out AND the browser never launched. This doesn't depend on `onTestEnd` running, so the single-in-flight-test case is handled. Verified against real Playwright (manual browser install, since the sandbox can't run `playwright install`): mid-test hang (browser launches, `goto` stalls) -> pw:api steps end err=0 -> browserReady true -> **no sentinel**; launch hang (fake sleeping browser) -> `Launch browser` never ends before onEnd -> browserReady false -> **sentinel**.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Also fixed by 6a8bed9d. The new signal is 'did the browser launch' (a successful `pw:api` step), not 'did a test complete', so a completed `timedOut`/`failed`/`passed` test no longer matters -- in all those cases the browser had launched, so browserReady is true and no sentinel is emitted.

probes/browser/templates/cloudprober-reporter.ts.tmpl
manugarg: Fixed in 6a8bed9d -- that comment (and the whole madeProgress block) is gone. The new comments describe the real pw:api/fixture lifecycle observed empirically, not the inaccurate 'failed with empty errors' claim.

Issue #1193, opened 2026-01-01
DNS probe for various authoritative dns servers


kerus1024:
I am currently using cloudprober in practice to monitor multiple DNS connectivity paths, including checks against specific DNS authoritative servers.

I previously used [smokeping](https://oss.oetiker.ch/smokeping/probe/AnotherDNS.en.html), but as the number of targets increased and the probing interval became shorter, the reliability of the generated RRD data degraded. In contrast, cloudprober, being a Go-based tool, does not fork a process per probe. By leveraging Prometheus increase() queries and latency buckets, it allows me to collect higher-resolution and more reliable metrics.

However, the current DNS probe implementation appears to always send queries with the RD (Recursion Desired) flag enabled. 
For my use case—especially when probing authoritative DNS servers—I need finer control over the DNS query behavior and response validation.

As a workaround, I am currently using the HTTP probe with a dedicated receiver server that performs DNS probing logic, similar to the following:

<details>

#### config and targets

```
probe {
  name: "dns-2s"
  type: HTTP
  interval_msec: 2100
  timeout_msec: 2100
  targets {
    file_targets {
      file_path: "./config/targets-dns-2s.yaml"
      re_eval_sec: 60
    }
  }
  stats_export_interval_msec: 5000

  http_probe {
    protocol: HTTP
    method: GET
    port: 2001
    user_agent: "cloudprober"
    latency_breakdown: FIRST_BYTE_LATENCY
  }

  validator {
      name: "status_code_2xx"
      http_validator {
          success_status_codes: "200"
      }
  }

  latency_unit: "ms"
  latency_distribution {
    explicit_buckets: "5,10,30,50,75,100,125,150,175,200,250,300,400,500,600,700,800,900,1000,1100,1200,1300,1400,1500,1600,1700,1800,1900,2000,INF"
  }

  ip_version: IPV4

  additional_label {
    key: "group"
    value: "@target.label.group@"
  }
  additional_label {
    key: "alert_group"
    value: "@target.label.alert_group@"
  }
  additional_label {
    key: "network"
    value: "@target.label.network@"
  }
  additional_label {
    key: "region"
    value: "@target.label.region@"
  }
  additional_label {
    key: "location"
    value: "@target.label.location@"
  }
  additional_label {
    key: "comment"
    value: "@target.label.comment@"
  }
}
```

```yaml
resource:
#
# Resolvers
#
- name: CloudFlare:1.1.1.1
  ip: 127.0.0.1
  labels:
[...]

PR #1396, opened 2026-07-28
[validators] Add DNS validator with authoritative answer check

manugarg:
## Summary
- New `dns_validator` with an `authoritative` criterion: response's AA flag must match the configured value. Useful for probing authoritative DNS servers (pairs with `recursion_desired: false` from #1395).
- DNS probe now passes the `*dns.Msg` to validators via `Input.Response`, like the HTTP probe does for HTTP responses. `Input.ResponseBody` still carries only the answer section, so existing regex validators are unaffected.

```
validator {
  name: "authoritative"
  dns_validator {
    authoritative: true
  }
}
```

Related: #1193

## Test plan
- `internal/validators/dns/dns_test.go`: Init (criteria required) and Validate (AA set/unset, both polarities, wrong input type).
- `internal/validators/validators_test.go`: `dns_validator` parses and inits through the oneof.
- `probes/dns/dns_test.go`: end-to-end - probe succeeds only when the mocked response is authoritative.
- `go test ./internal/validators/... ./probes/dns/...`


internal/validators/dns/proto/config.proto
```diff
@@ -0,0 +1,16 @@
...
+
+package cloudprober.validators.dns;
+
+option go_package = "github.com/cloudprober/cloudprober/internal/validators/dns/proto";
+
+// DNS validator configuration. It is applicable only to the DNS probe. For DNS
+// validator to succeed, all the conditions specified in the validator should
+// succeed.
+message Validator {
+  // If set to true, response should have the AA (Authoritative Answer) flag
```
manugarg: **[suggestion]** The `authoritative` field comment only describes the `true` case ("If set to true, response should have the AA flag set"), but the implementation and unit tests intentionally support both polarities: `authoritative: false` requires the AA flag to be unset. Readers of the proto may assume only `true` is meaningful.

**Suggestion:** Document both polarities, e.g. that the response AA flag must match the configured value (true ⇒ AA set, false ⇒ AA unset), and that the field must be set for the validator to be valid.

PR #1402, opened 2026-08-01
[release] Publish tar.gz archives and SHA-256 checksums

manugarg:
## Summary

- `make dist` now also builds `.tar.gz` archives for linux/macos (windows stays zip-only), so consumers don't need `unzip`.
- `make dist` emits `cloudprober-<version>-checksums.txt` with SHA-256 sums of every archive.
- The `tip` pre-release job renames all archives (not just zips) and regenerates checksums for the renamed files.

Groundwork for an `install.sh` (follow-up PR).

## Test plan

- Ran `make dist` locally: produced 6 zips + 5 tarballs + checksums; `sha256sum -c` passes; tarball contains `cloudprober-<version>-<platform>/cloudprober`.
- Simulated the `tip` rename + checksum regeneration on the real artifacts; `sha256sum -c` passes on the renamed set.

Note: for tagged releases, `cloudprober-<version>-checksums.txt` needs to be attached to the GitHub release along with the archives.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -294,7 +294,10 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: cloudprober-release-binaries
-          path: cloudprober-*.zip
+          path: |
+            cloudprober-*.zip
+            cloudprober-*.tar.gz
+            cloudprober-*-checksums.txt
 
   build_and_push_docker_multiarch:
     name: Build and push multiarch docker image
@@ -398,15 +401,17 @@ jobs:
         with:
           name: cloudprober-release-binaries
       - run: ls -R *
+      # Rename the versioned archives to "tip" ones, and regenerate checksums
+      # for the renamed files (the shipped checksums.txt refers to the
+      # versioned names).
       - run: |
-          mv cloudprober*-linux-arm64.zip cloudprober-tip-linux-arm64.zip
-          mv cloudprober*-linux-armv7.zip cloudprober-tip-linux-armv7.zip
-          mv cloudprober*-linux-x86_64.zip cloudprober-tip-linux-x86_64.zip
-          mv cloudprober*-macos-arm64.zip cloudprober-tip-macos-arm64.zip
-          mv cloudprober*-macos-x86_64.zip cloudprober-tip-macos-x86_64.zip
-          mv cloudprober*-windows-x86_64.zip cloudprober-tip-windows-x86_64.zip
+          rm -f cloudprober-*-checksums.txt
+          for f in cloudprober-*; do
+            mv "$f" "$(echo "$f" | sed -E 's/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/')"
+          done
+          sha256sum cloudprober-tip-* > cloudprober-tip-checksums.txt
       - uses: pyTooling/Actions/releaser@r0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           files: |
-            cloudprober-*.zip
+            cloudprober-tip-*
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -407,9 +407,15 @@ jobs:
       - run: |
           rm -f cloudprober-*-checksums.txt
           for f in cloudprober-*; do
-            mv "$f" "$(echo "$f" | sed -E 's/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/')"
+            tip=$(echo "$f" | sed -E 's/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/')
+            if [[ "$tip" == "$f" ]]; then
+              echo "Unexpected release artifact, can't derive a tip name: $f" >&2
+              exit 1
+            fi
+            mv "$f" "$tip"
           done
-          sha256sum cloudprober-tip-* > cloudprober-tip-checksums.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -398,15 +401,17 @@ jobs:
...
-          mv cloudprober*-linux-armv7.zip cloudprober-tip-linux-armv7.zip
-          mv cloudprober*-linux-x86_64.zip cloudprober-tip-linux-x86_64.zip
-          mv cloudprober*-macos-arm64.zip cloudprober-tip-macos-arm64.zip
-          mv cloudprober*-macos-x86_64.zip cloudprober-tip-macos-x86_64.zip
-          mv cloudprober*-windows-x86_64.zip cloudprober-tip-windows-x86_64.zip
+          rm -f cloudprober-*-checksums.txt
+          for f in cloudprober-*; do
+            mv "$f" "$(echo "$f" | sed -E 's/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/')"
+          done
+          sha256sum cloudprober-tip-* > cloudprober-tip-checksums.txt
```
manugarg: **[suggestion]** `sha256sum cloudprober-tip-* > cloudprober-tip-checksums.txt` expands a broad glob that also matches the checksums output path. On a clean first run this is usually fine (glob expands before the redirect creates the file), but it is a classic bash footgun: any re-run of the step, or any pre-existing `cloudprober-tip-checksums.txt`, can include the checksum file itself in the hashed set / produce confusing results. The Makefile side of this PR already avoids that by hashing only `*.zip` and `*.tar.gz`.

**Suggestion:** Mirror the Makefile and hash only archives, e.g. `sha256sum cloudprober-tip-*.zip cloudprober-tip-*.tar.gz > cloudprober-tip-checksums.txt`.

Makefile
```diff
@@ -70,9 +70,16 @@ dist: $(BINARIES)
...
-	  [[ "$${bin}" == *"windows"* ]] && mv $${bindir}/cloudprober{,.exe}; \
+	  if [[ "$${bin}" == *"windows"* ]]; then \
+	    mv $${bindir}/cloudprober{,.exe}; \
+	  else \
+	    tar -czf $${bindir}.tar.gz $${bindir}/; \
+	  fi; \
 	  zip -r $${bindir}.zip $${bindir}/; rm -rf $${bindir}; \
 	done
+	rm -f cloudprober-$(VERSION)-checksums.txt
+	sha256sum cloudprober-$(VERSION)-*.zip cloudprober-$(VERSION)-*.tar.gz \
```
manugarg: **[suggestion]** `sha256sum` is used unconditionally. That is fine on the Ubuntu `build-dist` runner, but `make dist` will fail on macOS developer machines (common for this repo’s dual linux/macos binary build), where the usual tool is `shasum -a 256` rather than `sha256sum`.

**Suggestion:** If local macOS `make dist` matters, use a small portable wrapper (e.g. prefer `sha256sum` when present, otherwise `shasum -a 256`) or document that `dist`/checksum generation is Linux-oriented. If dist is CI-only by design, this can be ignored.

.github/workflows/go.yml
```diff
@@ -398,15 +401,17 @@ jobs:
...
       - run: |
-          mv cloudprober*-linux-arm64.zip cloudprober-tip-linux-arm64.zip
-          mv cloudprober*-linux-armv7.zip cloudprober-tip-linux-armv7.zip
-          mv cloudprober*-linux-x86_64.zip cloudprober-tip-linux-x86_64.zip
-          mv cloudprober*-macos-arm64.zip cloudprober-tip-macos-arm64.zip
-          mv cloudprober*-macos-x86_64.zip cloudprober-tip-macos-x86_64.zip
-          mv cloudprober*-windows-x86_64.zip cloudprober-tip-windows-x86_64.zip
+          rm -f cloudprober-*-checksums.txt
+          for f in cloudprober-*; do
+            mv "$f" "$(echo "$f" | sed -E 's/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/')"
```
manugarg: **[nit]** The sed rename `s/^cloudprober-.*-(linux|macos|windows)-/cloudprober-tip-\1-/` is a nice generalization over the old hard-coded `mv` list, but if any unexpected `cloudprober-*` file does not match the pattern, `mv` becomes a same-file no-op and fails under GHA’s `set -e`, with a relatively opaque error.

**Suggestion:** Optionally guard the rename (only `mv` when the destination differs, or fail with an explicit message listing the unmatched file). Not blocking given the artifact set is tightly controlled.

.github/workflows/go.yml
manugarg: Done in 7faf225d — now `sha256sum cloudprober-tip-*.zip cloudprober-tip-*.tar.gz`, matching the Makefile.

Makefile
manugarg: Done in 7faf225d. Added `SHA256SUM := $(shell command -v sha256sum >/dev/null 2>&1 && echo sha256sum || echo "shasum -a 256")` and used that in `dist`. Verified both branches resolve correctly, and confirmed `shasum -a 256` emits the same `<hash>  <file>` format, so the checksums file parses identically either way.

.github/workflows/go.yml
manugarg: Done in 7faf225d — the loop now computes the tip name first and exits with `Unexpected release artifact, can't derive a tip name: <file>` if sed left it unchanged. This also covers the step being re-run, where every file is already `cloudprober-tip-*`.

PR #1403, opened 2026-08-01
[docs] Add install.sh and make it the primary install method

manugarg:
Stacked on #1402.

## Summary

- Adds `docs/static/install.sh`, served at `https://cloudprober.org/install.sh`. Detects platform, resolves the latest version from the `/releases/latest` redirect (no GitHub API rate limit), downloads the tarball, verifies its SHA-256, installs to `/usr/local/bin` or `~/.local/bin`. `VERSION` and `INSTALL_DIR` override the defaults.
- Getting Started leads with the one-liner; `go install` moves into the other-methods table.

## Test plan

- Served a fake release layout over localhost and ran the script end to end: downloads, verifies checksum, installs, installed binary reports the right `--version`.
- Failure paths: missing checksum file or checksum mismatch aborts before installing; missing archive gives a clear error.
- Verified the `/releases/latest` redirect resolves to `v0.14.4`.
- `sh -n` clean; confirmed `docs/static` is mounted to the site root in `docs/config/_default/module.toml`.

### DIFF
--- a/docs/static/install.sh
+++ b/docs/static/install.sh
@@ -110,8 +110,17 @@ curl -fsSL -o "${tmpdir}/${archive}" "${base_url}/${archive}" ||
 verify_checksum
 
 tar -xzf "${tmpdir}/${archive}" -C "$tmpdir"
-binary="${tmpdir}/cloudprober-${version}-${platform}/cloudprober"
-[ -f "$binary" ] || err "${archive} doesn't contain a cloudprober binary"
+# The archive holds a single directory, but its name tracks the build version
+# rather than $version ("tip" archives are renamed copies of versioned ones),
+# so find the binary instead of assuming the directory name.
+binary=""
+for f in "$tmpdir"/*/cloudprober; do
+  if [ -f "$f" ]; then
+    binary="$f"
+    break
+  fi
+done
+[ -n "$binary" ] || err "${archive} doesn't contain a cloudprober binary"
 
 pick_install_dir
 cp "$binary" "${install_dir}/cloudprober"
--- a/docs/static/install.sh
+++ b/docs/static/install.sh
@@ -1,0 +1,120 @@
+#!/bin/sh
+# Install script for cloudprober (https://cloudprober.org).
+#
+# Usage:
+#   curl -fsSL https://cloudprober.org/install.sh | sh
+#
+# Environment variables:
+#   VERSION      Release to install, e.g. v0.14.5. Defaults to the latest
+#                release. Use "tip" for the build from the main branch.
+#   INSTALL_DIR  Where to put the binary. Defaults to /usr/local/bin, falling
+#                back to $HOME/.local/bin if that's not writable.
+
+set -eu
+
+REPO_URL="https://github.com/cloudprober/cloudprober"
+
+err() {
+  echo "install.sh: $*" >&2
+  exit 1
+}
+
+# Map uname output to the platform names used in release archives.
+detect_platform() {
+  os=$(uname -s)
+  case "$os" in
+    Linux) os="linux" ;;
+    Darwin) os="macos" ;;
+    *) err "unsupported OS '$os'; see ${REPO_URL}/releases for other options" ;;
+  esac
+
+  arch=$(uname -m)
+  case "$arch" in
+    x86_64 | amd64) arch="x86_64" ;;
+    aarch64 | arm64) arch="arm64" ;;
+    armv7l | armv7) arch="armv7" ;;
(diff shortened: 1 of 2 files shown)


docs/static/install.sh
```diff
@@ -0,0 +1,130 @@
...
+
+echo "Downloading cloudprober ${version} (${platform})..."
+curl -fsSL -o "${tmpdir}/${archive}" "${base_url}/${archive}" ||
+  err "couldn't download ${base_url}/${archive}"
+
+verify_checksum
+
+tar -xzf "${tmpdir}/${archive}" -C "$tmpdir"
+# The archive holds a single directory, but its name tracks the build version
+# rather than $version ("tip" archives are renamed copies of versioned ones),
```
manugarg, line 128: **[suggestion]** Checksum lookup uses basic-regex `grep " ${archive}$"`. `archive` embeds `$version` from the environment, so a user-supplied `VERSION` containing regex metacharacters (e.g. `.` or `.*`) can match the wrong checksums line (or multiple lines), producing a confusing mismatch failure or, in pathological crafted files, a wrong expected hash. Official release names are safe, but fixed-string match is the right default for this comparison.

**Suggestion:** Use `grep -F` (and still anchor via shell checks if needed), e.g. match a line that ends with the literal archive name, or `awk -v a="$archive" '$2 == a || $NF == a { print $1; exit }'`.

docs/static/install.sh
```diff
@@ -0,0 +1,130 @@
...
+binary=""
+for f in "$tmpdir"/*/cloudprober; do
+  if [ -f "$f" ]; then
+    binary="$f"
+    break
+  fi
+done
+[ -n "$binary" ] || err "${archive} doesn't contain a cloudprober binary"
+
+pick_install_dir
```
manugarg, line 139: **[suggestion]** Install is a non-atomic `cp` of the extracted binary onto `${install_dir}/cloudprober`. An interrupted copy (full disk, kill -9, flaky remote FS) can leave a truncated/corrupt binary at the final path with no remaining temp artifact to recover from. `chmod` also runs only after `cp` returns, so a partial file may briefly exist with whatever mode `cp` created.

**Suggestion:** Copy to a same-directory temp name (e.g. `cloudprober.new.$$`), `chmod 755` that file, then `mv -f` over the destination so readers either see the old binary or the new one.

docs/static/install.sh
```diff
@@ -0,0 +1,130 @@
...
+tmpdir=$(mktemp -d)
+trap 'rm -rf "$tmpdir"' EXIT
+
+echo "Downloading cloudprober ${version} (${platform})..."
+curl -fsSL -o "${tmpdir}/${archive}" "${base_url}/${archive}" ||
+  err "couldn't download ${base_url}/${archive}"
+
+verify_checksum
+
+tar -xzf "${tmpdir}/${archive}" -C "$tmpdir"
```
manugarg, line 126: **[nit]** The PR description still frames missing checksums as “warn rather than fail” and the test plan says a missing checksums file “warns and continues.” The implementation hard-fails here (`refusing to install unverified`), which is the safer choice for a primary install method and already matches decision #2 in the desired end state. Stale PR text may cause reviewers to “fix” the wrong way or reintroduce soft-fail.

**Suggestion:** Update the PR body/test plan to match the hard-fail behavior and treat soft-fail as explicitly rejected (especially once any checksummed release exists). No code change required for this note.

docs/static/install.sh
manugarg: Fixed in 9ca286f7 — now `awk -v a="$archive" '$2 == a { print $1; exit }'`, an exact compare on the filename field.

Confirmed the old behavior against a two-line checksums file: with `VERSION='.*'` the BRE matched *both* lines and returned two concatenated hashes; awk returns empty for that name and the correct hash for the legitimate one.

docs/static/install.sh
manugarg: Fixed in 270bb374, along the lines you suggested — copy to `.cloudprober.$$` in the install dir, `chmod 755`, then `mv` into place, with the temp file cleaned up by the EXIT trap.

Worth noting this was not just theoretical: a plain `cp` onto a running cloudprober fails outright with `cp: cannot create regular file: Text file busy`, so re-running the installer while cloudprober was up would have errored. Verified the rename version succeeds against a live process.

docs/static/install.sh
manugarg: Already updated — the PR body now documents the hard-fail behavior and the test-plan table lists `checksums file missing (404) -> refusing to install unverified`. The stale text you saw predated 232d60c6. Agreed on treating soft-fail as explicitly rejected.

PR #1407, opened 2026-08-02
[validators] Reject unsupported validators

### LINKED ISSUE #1397: Reject probe-type-specific validators at config time
Probe-type-specific validators are silently inert when attached to the wrong probe type.

`http_validator` type-asserts its input to `*http.Response`, and the new `dns_validator` (#1396) asserts `*dns.Msg`. When the assertion fails, `Validate` returns an error, and `RunValidators` logs it and `continue`s:

https://github.com/cloudprober/cloudprober/blob/master/internal/validators/validators.go#L141-L144

So the validator never contributes a failure. A user who attaches `dns_validator { authoritative: true }` to an HTTP probe sees 100% success forever, with one log line per probe run. Same for `http_validator` on a non-HTTP probe. Related: validators are not supported at all for UDP / UDP_LISTENER probes, and configuring one there is silently ignored too.

This should be a config error at startup, not a runtime log.

### Suggested approach

Check it in `BuildProbeOptions`, next to the existing `negative_test` check, which is the same shape:

https://github.com/cloudprober/cloudprober/blob/master/probes/options/options.go#L199-L201

```go
var validatorProbeTypes = map[string][]configpb.ProbeDef_Type{
    "http_validator": {configpb.ProbeDef_HTTP},
    "dns_validator":  {configpb.ProbeDef_DNS},
}
```

`BuildProbeOptions` already has both halves in scope right where it calls `validators.Init`: `p.GetValidator()` (raw protos, oneof still intact) and `p.GetType()`.

Notes on why there rather than elsewhere:

- Not in each probe's `Init`: by then `opts.Validators` is `[]*validators.Validator`, i.e. just `{Name, Validate}` — `initValidator` has erased the oneof type, so a probe can't tell what it was handed without adding a type field to the struct. It would also mean every probe rejecting every foreign validator (N x M) instead of one table.
- Not in `validators.Init`: keeps the validator framework from having to know about probe types. `probes/options` already imports both.

The same table can cover the UDP case (no validators supported).

Guflly:
Rejects probe-specific validators when they are configured for incompatible probe types, instead of letting them fail silently at runtime.

The validation runs in `BuildProbeOptions` and covers HTTP, DNS, UDP, and UDP listener combinations.

Tests:
- `go test ./probes/options`
- `go vet ./probes/options`
- `go build ./probes/options`

Fixes #1397

### DIFF
--- a/probes/options/options.go
+++ b/probes/options/options.go
@@ -90,6 +90,46 @@ var negativeTestSupported = map[configpb.ProbeDef_Type]bool{
 	configpb.ProbeDef_HTTP: true,
 }
 
+var validatorProbeTypes = map[string][]configpb.ProbeDef_Type{
+	"http_validator": {configpb.ProbeDef_HTTP},
+	"dns_validator":  {configpb.ProbeDef_DNS},
+}
+
+var validatorsUnsupported = map[configpb.ProbeDef_Type]bool{
+	configpb.ProbeDef_UDP:          true,
+	configpb.ProbeDef_UDP_LISTENER: true,
+}
+
+func validateValidatorProbeTypes(p *configpb.ProbeDef) error {
+	if len(p.GetValidator()) == 0 {
+		return nil
+	}
+
+	if validatorsUnsupported[p.GetType()] {
+		return fmt.Errorf("validators are not supported by %s probes", p.GetType().String())
+	}
+
+	for _, validator := range p.GetValidator() {
+		if validator == nil {
+			continue
+		}
+
+		message := validator.ProtoReflect()
+		field := message.WhichOneof(message.Descriptor().Oneofs().ByName("type"))
+		if field == nil {
+			continue
+		}
+
+		validatorType := string(field.Name())
+		probeTypes, ok := validatorProbeTypes[validatorType]
+		if ok && !slices.Contains(probeTypes, p.GetType()) {
+			return fmt.Errorf("%s is not supported by %s probes", validatorType, p.GetType().String())
+		}
+	}
+
+	return nil
+}
+
 func defaultStatsExportInterval(p *configpb.ProbeDef, opts *Options) time.Duration {
 	minIntv := opts.Interval
 	if opts.Timeout > opts.Interval {
@@ -198,6 +238,10 @@ func ValidateProbeConfig(p *configpb.ProbeDef) (*Options, error) {
 
 	if p.GetNegativeTest() && !negativeTestSupported[p.GetType()] {
 		return nil, fmt.Errorf("negative_test is not supported by %s probes", p.GetType().String())
+	}
+
+	if err := validateValidatorProbeTypes(p); err != nil {
+		return nil, err
 	}
 
 	opts := &Options{
--- a/probes/options/options.go
+++ b/probes/options/options.go
@@ -28,6 +28,7 @@ import (
 	proberconfigpb "github.com/cloudprober/cloudprober/config/proto"
 	"github.com/cloudprober/cloudprober/internal/alerting"
 	"github.com/cloudprober/cloudprober/internal/validators"
+	validatorpb "github.com/cloudprober/cloudprober/internal/validators/proto"
 	"github.com/cloudprober/cloudprober/logger"
 	"github.com/cloudprober/cloudprober/metrics"
 	configpb "github.com/cloudprober/cloudprober/probes/proto"
@@ -100,6 +101,16 @@ var validatorsUnsupported = map[configpb.ProbeDef_Type]bool{
 	configpb.ProbeDef_UDP_LISTENER: true,
 }
 
+func validatorTypeName(v *validatorpb.Validator) string {
+	switch v.GetType().(type) {
+	case *validatorpb.Validator_HttpValidator:
+		return "http_validator"
+	case *validatorpb.Validator_DnsValidator:
+		return "dns_validator"
+	}
+	return ""
+}
+
 func validateValidatorProbeTypes(p *configpb.ProbeDef) error {
 	if len(p.GetValidator()) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


probes/options/options.go
```diff
@@ -90,6 +90,46 @@ var negativeTestSupported = map[configpb.ProbeDef_Type]bool{
...
+		return fmt.Errorf("validators are not supported by %s probes", p.GetType().String())
+	}
+
+	for _, validator := range p.GetValidator() {
+		if validator == nil {
+			continue
+		}
+
+		message := validator.ProtoReflect()
+		field := message.WhichOneof(message.Descriptor().Oneofs().ByName("type"))
```
manugarg: **[suggestion]** I'd rather not go through `Oneofs().ByName("type")` here. The oneof name is a magic string with no compile-time check, and `ByName` returns a nil descriptor on a miss while `WhichOneof(nil)` panics rather than returning nil - so a proto rename turns into a crash.

**Suggestion:** a plain type switch, mirroring what `initValidator` already does in internal/validators/validators.go, keeps everything compile-checked:

```go
func validatorTypeName(v *validatorpb.Validator) string {
	switch v.GetType().(type) {
	case *validatorpb.Validator_HttpValidator:
		return "http_validator"
	case *validatorpb.Validator_DnsValidator:
		return "dns_validator"
	}
	return ""
}
```

The loop then becomes a plain name lookup against `validatorProbeTypes`, and an unmatched/unset type falls out as `""`, which the map won't contain.

probes/options/options.go
```diff
@@ -90,6 +90,46 @@ var negativeTestSupported = map[configpb.ProbeDef_Type]bool{
...
+	}
+
+	if validatorsUnsupported[p.GetType()] {
+		return fmt.Errorf("validators are not supported by %s probes", p.GetType().String())
+	}
+
+	for _, validator := range p.GetValidator() {
+		if validator == nil {
+			continue
+		}
```
manugarg: **[nit]** This nil check isn't reachable - proto unmarshaling doesn't produce nil entries in a repeated message field, and if one somehow existed, `validators.Init` would panic on `vc.Name` a few lines later anyway. Can be dropped.

probes/options/options.go
```diff
@@ -90,6 +90,46 @@ var negativeTestSupported = map[configpb.ProbeDef_Type]bool{
...
+		message := validator.ProtoReflect()
+		field := message.WhichOneof(message.Descriptor().Oneofs().ByName("type"))
+		if field == nil {
+			continue
+		}
+
+		validatorType := string(field.Name())
+		probeTypes, ok := validatorProbeTypes[validatorType]
+		if ok && !slices.Contains(probeTypes, p.GetType()) {
+			return fmt.Errorf("%s is not supported by %s probes", validatorType, p.GetType().String())
```
manugarg: **[suggestion]** With several validators configured, this doesn't say which one is at fault. `validator.GetName()` is right here:

```go
return fmt.Errorf("validator %q: %s is not supported by %s probes", validator.GetName(), validatorType, p.GetType().String())
```

probes/options/options_test.go
```diff
@@ -674,6 +675,64 @@ func TestValidateProbeConfig(t *testing.T) {
...
+		{
+			name: "validators_unsupported_udp_listener",
+			probe: &configpb.ProbeDef{
+				Type: configpb.ProbeDef_UDP_LISTENER.Enum(),
+				Validator: []*validatorpb.Validator{{
+					Type: &validatorpb.Validator_Regex{Regex: "ok"},
+				}},
+			},
+			wantErr: "validators are not supported by UDP_LISTENER probes",
+		},
```
manugarg, line 782: **[suggestion]** Allowed paths for generic validators are not covered. There is no case that asserts `regex` / `json_validator` / `integrity_validator` still succeed on HTTP, DNS, or other non-UDP probe types. A future change that accidentally rejects unknown/generic oneof names (or flips the map logic) would not be caught.

**Suggestion:** Add one or two positive cases, e.g. HTTP + `regex` and DNS + `regex` (and optionally HTTP + `json_validator`), with no `wantErr`.

Two more negative controls worth adding, both of which pass today even if the check is broken:

- A `UDP` probe with **no** validators should succeed - nothing currently guards the `len(p.GetValidator()) == 0` early return; delete that early return and every test still passes.
- A case with two validators where the *second* one is the bad one, so the loop itself is covered rather than just the first element.

probes/options/options.go
Guflly: Removed it.

probes/options/options.go
Guflly: Added the validator name to the error.

probes/options/options_test.go
Guflly: Added generic-validator cases, UDP without validators, and a case where the second validator is invalid.

probes/options/options.go
Guflly: Switched this to a type switch.

probes/options/options.go
```diff
@@ -90,6 +91,46 @@ var negativeTestSupported = map[configpb.ProbeDef_Type]bool{
...
+
+func validatorTypeName(v *validatorpb.Validator) string {
+	switch v.GetType().(type) {
+	case *validatorpb.Validator_HttpValidator:
+		return "http_validator"
+	case *validatorpb.Validator_DnsValidator:
+		return "dns_validator"
+	}
+	return ""
+}
```
manugarg: I think this indirection is unncessary. `validatorTypeName` converts the validator's oneof case to a string, which only exists so `validateValidatorProbeTypes` can turn around and use that string as a key into `validatorProbeTypes`. That's two data structures (a switch and a map) that have to be kept in sync by hand, with no compiler check tying them together.

Collapse both into a single function that switches on the oneof once and returns everything the caller needs:

```go
func validatorProbeTypes(v *validatorpb.Validator) (name string, types []configpb.ProbeDef_Type) {
	switch v.GetType().(type) {
	case *validatorpb.Validator_HttpValidator:
		return "http_validator", []configpb.ProbeDef_Type{configpb.ProbeDef_HTTP}
	case *validatorpb.Validator_DnsValidator:
		return "dns_validator", []configpb.ProbeDef_Type{configpb.ProbeDef_DNS}
	}
	return "", nil
}
```

manugarg (commented):
Logic and tests look solid, but the `validatorTypeName` + `validatorProbeTypes` split adds a layer of indirection that isn't earning its keep. Please collapse it before merging — see inline comment.

manugarg:
Thanks for the fix, @Guflly! Since this had been idle for a while, I pushed the last review change (collapsing `validatorTypeName` and the `validatorProbeTypes` map into a single switch) directly onto the branch so we can get it merged.

PR #1410, opened 2026-08-12
Added an option to opt out of logging query parameters

markoposavec:
Hello,

I've added an opt-in redact_url_query_in_logs field to the HTTP probe config (default false, so existing behavior is unchanged); when enabled, URL query parameters are replaced with "\<redacted>" in probe logs.

Some of our legacy endpoints carry sensitive data in query parameters, and when a probe request fails cloudprober logs the full request URL including the query string, so those secrets end up in the logs with no way to opt out short of disabling logging entirely.

Also, in my opinion, probe validation failures shouldn't produce error logs: the system (cloudprober) is working properly - the probe is failing, which isn't a fault of the system itself, so surfacing it at Error feels misleading. But that's really a separate discussion. :)

### DIFF
--- a/probes/http/http.go
+++ b/probes/http/http.go
@@ -63,9 +63,10 @@ type Probe struct {
 	redirectFunc  func(req *http.Request, via []*http.Request) error
 
 	// book-keeping params
-	targets []endpoint.Endpoint
-	method  string
-	url     string
+	targets              []endpoint.Endpoint
+	method               string
+	url                  string
+	redactURLQueryInLogs bool
 	// Canonical names of headers whose values change under substitution
 	// (e.g. @uuid@). Empty when no header is dynamic; len() also gates the
 	// per-send clone in prepareRequest.
@@ -199,6 +200,7 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 	if len(p.url) > 0 && p.url[0] != '/' {
 		return fmt.Errorf("invalid relative URL: %s, must begin with '/'", p.url)
 	}
+	p.redactURLQueryInLogs = p.c.GetRedactUrlQueryInLogs()
 
 	body := p.c.GetBody()
 	if len(body) == 0 && p.c.GetBodyFile() != "" {
@@ -309,7 +311,7 @@ func (p *Probe) requestTrace(result *probeResult) *httptrace.ClientTrace {
 
 // doHTTPRequest executes an HTTP request and updates the provided result struct.
 func (p *Probe) doHTTPRequest(req *http.Request, client *http.Client, target endpoint.Endpoint, result *probeResult, resultMu *sync.Mutex) error {
-	l := p.l.WithAttributes(slog.String("target", target.Name), slog.String("url", req.URL.String()))
+	l := p.l.WithAttributes(slog.String("target", target.Name), slog.String("url", p.redactedURL(req.URL)))
 
 	start := time.Now()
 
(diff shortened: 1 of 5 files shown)


manugarg:
This mostly LGTM. I'll take a deeper look soon.

manugarg (commented):
Single-run/`RunOnce` leaks the query: `doHTTPRequest` returns the raw `err` (a `*url.Error` with the full URL), which eventually gets printed unredacted. Suggest returning a redacted error (e.g. `errors.New(p.redactedErr(req.URL, err))`) at http.go:342/354.

Worth gating the rewrap on the flag, so the default path keeps the typed error:

```go
if p.redactURLQueryInLogs {
	return errors.New(p.redactedErr(req.URL, err))
}
return err
```

Rewrapping unconditionally is safe today -- `isClientTimeout` does its `*url.Error` assertion at http.go:334, before the return, and nothing downstream of `runProbe` type-asserts on this error -- but leaving `err` untouched when the option is off keeps the "default false, existing behavior unchanged" contract exact.

probes/http/request.go
```diff
@@ -96,6 +97,35 @@ func pathForTarget(target endpoint.Endpoint, probeURL string) string {
...
+// query via the error message.
+func (p *Probe) redactedErr(u *url.URL, err error) string {
+	if err == nil {
+		return ""
+	}
+	msg := err.Error()
+	if !p.redactURLQueryInLogs || u.RawQuery == "" {
+		return msg
+	}
+	return strings.ReplaceAll(msg, u.String(), p.redactedURL(u))
```
manugarg: Could match on `"?"+u.RawQuery` instead of the full `u.String()` -- same result, but it doesn't break when something rewrites the rest of the URL, and it'd work at the `http.go:602` call site where there's no clean `*url.URL` to hand it.

The concrete case for "rewrites the rest of the URL" is a redirect: `client.Do` follows it, the later hop fails, and `net/http` builds the `*url.Error` from the *redirected* request, so `u.String()` matches nothing. An HTTP->HTTPS upgrade redirect preserves the query, so it's the common shape, not a corner case:

| | `u.String()` | `"?"+u.RawQuery` |
| --- | --- | --- |
| plain failure | redacted | redacted |
| redirect to another host, query preserved | **leaks** | redacted |
| fragment in URL | redacted | redacted |
| `"` in query | **leaks** | **leaks** |

Two caveats:

1. Neither form handles a quote in the query. `url.Error.Error()` renders the URL with `%q`, so `sig=x"y` appears in the error text as `sig=x\"y` and no substring match hits. Narrow enough to leave alone, but if you want it closed it needs a second pass over the `strconv.Quote`d query.
2. For the http.go:602 site, the raw query has to come from somewhere. `p.url` only covers the probe-level `relative_url` -- targets can supply their own path (and query) via the `relative_url` label or `__cp_path__`, see `pathForTarget`, so caching `p.url`'s query at Init would silently miss those.

manugarg (commented):
`http.go:602` leaks the same way in single-run mode -- `httpRequestForTarget`'s error is logged verbatim and also passed to `LastRun.Set`. With `relative_url: "/login?password=secret\x7f"` it produces:

```
error creating HTTP request: parse "http://t1/login?password=secret\x7f": net/url: invalid control character in URL
```

Worth covering in the same pass.

probes/http/http.go
```diff
@@ -199,6 +200,7 @@ func (p *Probe) Init(name string, opts *options.Options) error {
 	if len(p.url) > 0 && p.url[0] != '/' {
 		return fmt.Errorf("invalid relative URL: %s, must begin with '/'", p.url)
 	}
+	p.redactURLQueryInLogs = p.c.GetRedactUrlQueryInLogs()
```
manugarg: `Init` leaks the query too, and the flag is assigned two lines below the check that leaks it. The validation error at http.go:201 echoes the raw `relative_url`, so:

```
relative_url: "login?password=secret"   # missing leading '/'
```

fails startup with `invalid relative URL: login?password=secret, must begin with '/'`, secret intact, regardless of how `redact_url_query_in_logs` is set.

Moving this assignment above the `p.url[0] != '/'` check and redacting the message would close it -- and it's a good argument for the redaction helper taking a raw query string rather than a `*url.URL`, since there's no parsed URL here either.

probes/http/request.go
manugarg: Fixed.

probes/http/http.go
manugarg: Fixed.

manugarg:
@markoposavec I took the liberty to follow up with the review fixes. Hope that's okay with you. Please let me know if it's not, I'll avoid it next time.

PR #1414, opened 2026-08-18
docs: document the file surfacer output format

### LINKED ISSUE #18: [Documentation] Document file surfacer output
This issue started when the repository was under Google, see full context here: google/cloudprober#633

I'm setting up a probing configuration that uses Cloudprober. I'm looking at Cloudprober output which looks similar to this:

```
cloudprober 15.. 1500590520 labels=ptype=http,probe=google-http,dst=.. total=17 success=17 latency=180835
cloudprober 15.. 1500590530 labels=ptype=sysvars,probe=sysvars hostname="manugarg-ws" uptime=100
cloudprober 15.. 1500590530 labels=ptype=http,probe=google-http,dst=.. total=19 success=19 latency=211644
```

I want to be able to read every line and understand every word of it. There should be documentation that explains it.

The specific output above is from the docs and is mostly self-explanatory, but the examples I've got (which I can't share) aren't as easy.

vshulcz:
Closes #18.

The file surfacer's output format wasn't documented anywhere, so the only way to read a line was to work back from `EventMetrics.String()`. Adds a page for it and links it from the surfacers overview next to stackdriver and cloudwatch.

It covers the four parts of a line (prefix, id, timestamp, labels, metrics), and the two value shapes that aren't plain numbers: maps as `map:code,200:44,204:8` and distributions as `dist:sum:899|count:221|lb:...|bc:...`, including how the bucket bounds line up with the counts.

Everything is taken from the code rather than from example output: `internal/surfacers/file/file.go` for the line layout and the id seeded from the nanosecond timestamp, `metrics/eventmetrics.go` for the labels and metrics sections, and `metrics/map.go`, `metrics/dist.go`, `metrics/string.go` for the value formats. The `us` default for `latency_unit` comes from `probes/proto/config.proto`.

### DIFF
--- a/docs/content/docs/surfacers/file.md
+++ b/docs/content/docs/surfacers/file.md
@@ -1,0 +1,87 @@
+---
+title: "File Surfacer"
+menu:
+  docs:
+    parent: "surfacers"
+    weight: 25
+---
+
+The file surfacer writes metrics as lines of text, either to a file or, if no
+file path is configured, to standard output. It's enabled automatically, along
+with the prometheus surfacer, if you don't configure any surfacer at all.
+
+## Output format
+
+Each line corresponds to one
+[EventMetrics](https://pkg.go.dev/github.com/cloudprober/cloudprober/metrics#EventMetrics)
+record and is built from four parts:
+
+```
+cloudprober 1500 1500590520 labels=ptype=http,probe=google-homepage,dst=google.com total=17 success=17 latency=180835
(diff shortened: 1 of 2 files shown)


manugarg:
@vshulcz are you using file surfacer for something?

File surfacer is there mainly because of legacy reasons, I don't really see people using it anymore. That's the reason it's not documented.

vshulcz:
Not in production. I came from #18, which is still open and carried over from google/cloudprober#633, so it looked undocumented rather than intentionally undocumented.

What made me write it up: the surfacers overview says prometheus and file are enabled automatically when no surfacer is configured, so these lines are what a new user sees first even if they never pick the file surfacer.

If you'd rather keep the docs lean, I can cut this to a couple of sentences in the overview, or close it and let #18 go as won't do. Your call.

PR #1424, opened 2026-08-19
[rds.kubernetes] Add HTTPRoute (Gateway API) target discovery

### LINKED ISSUE #1409: Kubernetes target discovery: support for the Gateway API (HTTPRoute resource)
Unless I missed something, there's currently no support for discovering `HTTPRoute` Kubernetes resources. 

These are part of the [Gateway API](https://kubernetes.io/docs/concepts/services-networking/gateway/), an alternative to the Ingress API (v1 in [October 2023](https://kubernetes.io/blog/2023/10/31/gateway-api-ga/)).

We migrated some of our Ingress to HTTPRoute objects and cloudprober target discovery doesn't seem to support them.

Is that something that would be in scope for `cloudprober`?

manugarg:
Adds Kubernetes target discovery for Gateway API `HTTPRoute` resources, analogous to the existing `Ingress` support. Closes #1409.

## What it does
- New `http_routes` resource type for the `k8s` targets provider.
- Each `HTTPRoute` is expanded into one RDS resource per (hostname x match path), mirroring how ingresses are expanded:
  - `name`: `<route>_<host>_<path>`
  - `fqdn` label: hostname
  - `relative_url` label: match path (`/` for rules without a path match)
  - `namespace` label, as for the other kinds (#1489)
- Hostnames come from `spec.hostnames` (Gateway API has no rule-level hostnames).
- Routes without hostnames (they inherit them from the Gateway listener) and wildcard hostnames (e.g. `*.example.com`) produce no resources, since there is no specific host to probe.
- `RegularExpression` path matches are skipped, as they don't name a specific URL.
- If the route has its own `relative_url` label, paths are not expanded: one resource per hostname, named `<route>_<host>`.
- Rules that expand to the same (hostname, path), e.g. rules differing only in header or method matches, produce a single resource.

## Note on the target IP
Unlike `Ingress`, an `HTTPRoute` does not carry a load balancer IP in its status (the address lives on the parent `Gateway`). This change uses the route hostname as the target IP; the probe resolves it via DNS, and the `fqdn` label drives the Host header / SNI. Resolving the parent `Gateway.status.addresses` for a real IP is a possible follow-up, as is using the route's status conditions as a readiness signal.

Requires the Gateway API CRDs to be installed in the cluster.

## Config
```
targets {
  k8s {
    http_routes: ".*"
  }
}
```

## Tests
- `httproutes_test.go` + `testdata/httproutes.json` cover parsing, listing, and name/label/namespace filtering, skipping of hostname-less and wildcard routes and regex paths, the `relative_url` label override, and deduplication of identical (hostname, path) pairs.
- `targets/k8s_test.go` covers the `http_routes` config.

### DIFF
--- a/docs/content/docs/how-to/k8s_targets.md
+++ b/docs/content/docs/how-to/k8s_targets.md
@@ -51,7 +51,9 @@ unlike `Ingress`). The probe resolves it via DNS, and the hostname is also
 exposed as the `fqdn` label so HTTP probes set the correct Host header / SNI.
 Routes without hostnames (which inherit them from the Gateway listener) and
 wildcard hostnames (e.g. `*.example.com`) are skipped, as there is no
-specific host to probe. HTTPRoute discovery requires the
+specific host to probe, and so are `RegularExpression` path matches. If the
+route has its own `relative_url` label, paths are not expanded and you get one
+target per hostname. HTTPRoute discovery requires the
 [Gateway API](https://gateway-api.sigs.k8s.io/) CRDs to be installed in the
 cluster.
 
@@ -116,8 +118,8 @@ You can filter k8s resources using the following options:
 Discovered targets carry the Kubernetes resource's labels, plus a `namespace`
 label set to the resource's namespace. Endpoints targets also get a `node`
 label, and a `pod` label if the address belongs to a pod. Ingress and HTTPRoute
-targets get `fqdn` and `relative_url` labels. If the resource already has a label with one
-of these names, its value is kept.
+targets get `fqdn` and `relative_url` labels. If the resource already has a
+label with one of these names, its value is kept.
 
 You can use these labels in
 [additional labels]({{< ref "additional-labels.md" >}}), e.g. to add the
--- a/docs/content/docs/how-to/k8s_targets.md
+++ b/docs/content/docs/how-to/k8s_targets.md
@@ -43,6 +43,12 @@ Cloudprober supports discovery for the following k8s resources:
 - **Endpoints**
 - **Pods**
 - **Ingresses**
+- **HTTPRoutes** (Gateway API)
+
+Note: For `HTTPRoute` resources, the discovered target IP is the route's
+hostname (an `HTTPRoute` does not carry a load balancer IP in its status,
+unlike `Ingress`). The probe resolves it via DNS, and the hostname is also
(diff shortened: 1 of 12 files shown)


wingsofovnia:
Thank you for this contribution! Support of `GRPCRoute` would also be great. It's the same exact structure (for what is relevant here) 🙏

manugarg:
Sorry for the delay here. I was waiting for the refactor that I just merged through #1486 before adding more resource types.

manugarg:
Rebased onto `main` now that #1486 (generic `resourceLister`) and #1485 have landed, and converted the lister to the new pattern.

`httproutes.go` goes from 253 to 152 lines. What went away was the fifth copy of the shared boilerplate — the `httpRoutesLister` struct, `httpRoutesURL`, `listResources`, `parseHTTPRoutesJSON`, `expand`, and the refresh goroutine. What stayed is the part that's actually about HTTPRoute: the route types and `resources()`.

Worth noting the rebase itself was clean and the result compiled and passed tests **with the old pattern still in place** — the additions never textually overlapped what #1486 rewrote. So nothing would have flagged that this was still a fifth copy; the conversion had to be deliberate.

Three things it inherits for free:

- `expand()` keeps the last good result when a refresh fails. The version in this PR had the same fall-through that emptied the cache on any API error.
- The refresh loop no longer calls the deprecated `rand.Seed`, and the stale "gceInstances / GCE API" comment is gone with it.
- A name in the resource path (`k8s://httproutes/my-route`) selects a single route, consistent with every other type.

**One judgment call worth a look.** HTTPRoute filters name and labels on the *expanded* resources rather than on the route object, which is what this PR already did and matches ingresses — derived names and per-resource `fqdn`/`relative_url` labels would be unselectable otherwise.

The subtlety is the fallback resource for a route with no hostnames. It has to be chosen *before* filtering, otherwise a route whose resources are all filtered out reappears under its bare name. The naive conversion gets this wrong, and this PR's own tests catch it — `name filter for host regex` picks up a spurious `rule-host-route`. Kept the original ordering and added a comment saying why.

Also added httproutes to `TestListerAPIPaths`, so the gateway.networking.k8s.io prefix is pinned like the other four. Package coverage 83.9% -> 84.8%. `go build ./...` and `go test -race ./internal/rds/... ./targets/...` pass.

Issue #1431, opened 2026-08-27
HTTP/HTTPS file target generates errors and does not work


Synehan:
**Describe the bug**
When using file target with http (or https) path, Cloudprober return an error

**Cloudprober Version**
v0.14.5

**To Reproduce**
1. Create a probe with file_target target and an http endpoint.

```
probe {
    name: "all-endpoints"
    type: HTTP

    additional_label {
        key: "cluster"
        value: "@target.label.dc@"
    }

    targets {
        file_targets {
            file_path: "https://raw.githubusercontent.com/cloudprober/cloudprober/refs/heads/main/examples/targets/file/targets.textpb"
            re_eval_sec: 300
        }
    }
}
```
2. Launch cloudprober et look at the logs.
```
level=ERROR source=internal/rds/file/file.go:257 msg="file(https://raw.githubusercontent.com/cloudprober/cloudprober/refs/heads/main/examples/targets/file/targets.textpb): error while reading file: Get \"raw.githubusercontent.com/cloudprober/cloudprober/refs/heads/main/examples/targets/file/targets.textpb\": unsupported protocol scheme \"\"" system=cloudprober probe=all-endpoints
```

**Additional context**
Problem is caused by the strip of the schema for all file remote file type (s3/gcs/http/https). This ok for gcs and s3 as the result is use as the objectPath but http does not work. The golang lib net/http need the scheme to work.

PR #1432, opened 2026-08-27
[internal.file] Don't strip the scheme from http(s):// file paths

Synehan:
Addresses #1431. 

Stripping http:// and https:// with http.NewRequestWithContext generate error and makes file_target with http not working.

I have deported the stripping into each s3 and gcs functions.

I have also moved http functions to a specific file.

Build / Docker tested + live tested

### DIFF
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -20,8 +20,6 @@ package file
 import (
 	"context"
 	"fmt"
-	"io"
-	"net/http"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


internal/file/file.go
```diff
@@ -149,7 +103,7 @@ func ReadFile(ctx context.Context, fname string, readOptions ...ReadOption) ([]b
 
 	for prefix, f := range prefixToReadfunc {
 		if strings.HasPrefix(fname, prefix) {
-			return processContent(f(ctx, fname[len(prefix):]))
+			return processContent(f(ctx, fname))
```
manugarg, line 106: if you added added a simple check here to not strip prefix if prefix is http or https, that would avoid all other changes. This + test would be sufficient and will make it easy to compare and review.

Even if you didn't want this new condition, to keep the logic same for all, it will be a few lines change, again making it easier to review.

manugarg: More concrete suggestion:
- Don't do the http.go split. It makes reviewing much harder.
- Add a function called ` func trimPrefixIfNeeded(prefix, fname) string` and pass that to processContent. Also change ModTime.
- trimPrefixIfNeeded could skip prefixing for http/https prefixes.

WDYT?

manugarg: Since it's been a while, I pushed a commit to the same branch. Hope that's okay.

PR #1457, opened 2026-09-02
[ci] Clean stale fingerprinted assets from the gh-pages tree

manugarg:
Hugo rewrites its output tree in place and never deletes fingerprinted assets it no longer emits. Every time an asset's content (and therefore its fingerprint) changes — e.g. the search bundle, which embeds the site content — a new `index.min.<sha>.js` / `main.<sha>.css` / `*_hu<sha>_*.webp` lands in the tree while the old one is kept. The gh-pages branch has accumulated **174 stale bundles (~63 MiB, tree is 92 MiB total)** this way.

## What

- `tools/clean_stale_assets.sh`: deletes fingerprint-named files in the built tree that no page references (scans html/css/js/json/xml; supports `-n`/`--dry-run` in any argument position). Fails closed: a reference scan that errors (grep exit 2), or finds zero references while candidates exist, aborts the run instead of deleting the asset set.
- Both site-deploy workflows (`hugo.yml`, `latest_config_docs.yml`) run it after the build, so the next deploy commit also carries the deletions.

## Why reference-based, not "wipe and rebuild"

The versioned config docs (`docs/config/v0.13.7` … `v0.14.5`) exist **only** as committed artifacts in the gh-pages tree — a normal docs push only regenerates `config/main` + `config/latest`, so wiping the tree would delete them. Those surviving pages also still reference the bundles they were built with (e.g. 117 pages under `v0.14.5` pin one old `index.min`), so cleanup must keep anything referenced from anywhere in the tree. The script only ever touches fingerprint-named files, and only when unreferenced, so HTML pages and non-fingerprinted files (`configdoc.js`, `CNAME`, …) are out of scope by construction.

## Verified against clones of the branches

- Dry run on current gh-pages: 174 stale assets found; exactly the referenced files kept
- Real run: 62.9 MiB freed (92M → 30M), second run is a no-op, versioned config docs / `configdoc.js` / `CNAME` / `robots.txt` untouched
- Simulated a page being rebuilt with the current bundle: the next run frees the old bundle it pinned
- Repro from review (fingerprinted files but no HTML at all): aborts with an error, deletes nothing

## Residue note

Bundles referenced by pages that are not rebuilt stay alive — that is intentional, but it only "drains" for pages that do get rebuilt (versioned config docs, via the `latest_config_docs.yml` dispatch). Pages in fully orphaned directories pin bundles forever; `docs/config/master/` (pre-branch-rename fossil, zero inbound links, not in the sitemap, not enumerable by the version selector) was removed from gh-pages as a one-off (`8230d999`), which freed the two bundles it exclusively pinned. Other unreachable top-level sections (`concepts/`, `getting-started/`, `surfacers/`, `how-to/`, `goto/`) remain: several are still link targets of current pages via stale URLs, so they need a content fix before removal — follow-up.

### DIFF
--- a/tools/clean_stale_assets.sh
+++ b/tools/clean_stale_assets.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Delete fingerprinted assets that no page in the built site references.
+#
+# Hugo rewrites its output tree in place and never deletes files it no
+# longer emits, so every time an asset's content (and therefore its
+# fingerprint) changes -- e.g. the search bundle, which embeds the site
+# content -- a new index.min.<sha>.js / main.<sha>.css / *_hu<sha>_....webp
+# lands in the tree while the old one is kept. Over time that accumulates
+# hundreds of stale bundles in the gh-pages branch.
+#
+# Only fingerprint-named files are ever considered, and only when
+# unreferenced, so files that live on in the tree without being rebuilt by
+# this run -- in particular the versioned config docs, which reference the
+# bundles they were built with -- keep working.
+
+set -euo pipefail
+
+DRY_RUN=0
+if [[ "${1:-}" == "-n" ]]; then
+  DRY_RUN=1
+  shift
+fi
+DIR="${1:?usage: clean_stale_assets.sh [-n] <built-site-dir>}"
+
+cd "$DIR"
+
+# Fingerprint patterns: the asset pipeline emits <name>[.min].<sha>.js|css,
+# the image pipeline <name>_hu<sha>_<params>.webp|png|jpg|avif.
+FP='[A-Za-z0-9_/.-]*\.[0-9a-f]{64,}\.(js|css)|[A-Za-z0-9_/.-]*_hu[0-9a-f]{32}_[A-Za-z0-9_]*\.(webp|png|jpe?g|avif)'
+
+# Basenames referenced by any built file. HTML is minified with unquoted
+# attributes (src=https://...), so match bare filename patterns instead of
+# attribute syntax. The search bundle embeds page content, so scanning the
+# JS as well covers references from pages built in this run.
+referenced=$(grep -rhoE "$FP" \
+  --include='*.html' --include='*.css' --include='*.js' \
+  --include='*.json' --include='*.xml' . | sed 's|.*/||' | sort -u || true)
+
+deleted=0
+freed=0
+while IFS= read -r f; do
+  base=${f##*/}
+  if ! grep -qxF "$base" <<<"$referenced"; then
+    if [[ "$DRY_RUN" == 1 ]]; then
+      echo "would delete: $f"
+    else
+      freed=$((freed + $(stat -c%s "$f")))
+      rm "$f"
+      echo "deleted: $f"
+    fi
+    deleted=$((deleted + 1))
+  fi
+done < <(find . -path ./.git -prune -o -type f -print |
+  grep -E '\.[0-9a-f]{64,}\.(js|css)$|_hu[0-9a-f]{32}_[A-Za-z0-9_]+\.(webp|png|jpe?g|avif)$' || true)
+
+if [[ "$DRY_RUN" == 1 ]]; then
+  echo "dry run: $deleted stale assets found"
+else
+  echo "deleted $deleted stale assets ($((freed / 1024)) KiB)"
+fi
--- a/tools/clean_stale_assets.sh
+++ b/tools/clean_stale_assets.sh
@@ -16,26 +16,63 @@
 
 set -euo pipefail
 
+usage() {
+  echo "usage: clean_stale_assets.sh [-n|--dry-run] <built-site-dir>" >&2
+  exit 2
+}
+
 DRY_RUN=0
-if [[ "${1:-}" == "-n" ]]; then
-  DRY_RUN=1
-  shift
-fi
-DIR="${1:?usage: clean_stale_assets.sh [-n] <built-site-dir>}"
+DIR=""
+for arg in "$@"; do
+  case "$arg" in
+    -n|--dry-run) DRY_RUN=1 ;;
+    -h|--help) usage ;;
+    -*) echo "error: unknown flag: $arg" >&2; usage ;;
+    *)
+      [[ -z "$DIR" ]] || { echo "error: unexpected argument: $arg" >&2; usage; }
+      DIR="$arg"
+      ;;
+  esac
+done
+[[ -n "$DIR" ]] || usage
+[[ -d "$DIR" ]] || { echo "error: not a directory: $DIR" >&2; exit 2; }
 
 cd "$DIR"
 
 # Fingerprint patterns: the asset pipeline emits <name>[.min].<sha>.js|css,
 # the image pipeline <name>_hu<sha>_<params>.webp|png|jpg|avif.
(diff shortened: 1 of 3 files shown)


tools/clean_stale_assets.sh
```diff
@@ -0,0 +1,61 @@
...
+# the image pipeline <name>_hu<sha>_<params>.webp|png|jpg|avif.
+FP='[A-Za-z0-9_/.-]*\.[0-9a-f]{64,}\.(js|css)|[A-Za-z0-9_/.-]*_hu[0-9a-f]{32}_[A-Za-z0-9_]*\.(webp|png|jpe?g|avif)'
+
+# Basenames referenced by any built file. HTML is minified with unquoted
+# attributes (src=https://...), so match bare filename patterns instead of
+# attribute syntax. The search bundle embeds page content, so scanning the
+# JS as well covers references from pages built in this run.
+referenced=$(grep -rhoE "$FP" \
+  --include='*.html' --include='*.css' --include='*.js' \
+  --include='*.json' --include='*.xml' . | sed 's|.*/||' | sort -u || true)
```
manugarg: **An empty reference set silently deletes every fingerprinted asset.**

The reference scan ends in `|| true`, so any failure is swallowed and `referenced` comes back empty or partial: grep exiting 2 on a read error, a truncated `public/` from a partially failed Hugo build, or a future edit that breaks the `--include` list or `$FP`. Line 44 then treats *every* candidate as unreferenced.

I reproduced this: on a directory containing only `main.<64hex>.css` and `index.min.<64hex>.js` and no HTML, the script reports both as stale and would delete them, exiting 0. In CI that means the Deploy step commits the removal of all 184 fingerprinted bundles/images and pushes it — and because a legitimate first run already deletes 171 files, the extra 13 blend into the diff.

Suggest aborting when `referenced` is empty while candidates exist, and dropping the blanket `|| true` (or distinguishing grep's exit 1 from exit 2).

tools/clean_stale_assets.sh
```diff
@@ -0,0 +1,61 @@
...
+
+deleted=0
+freed=0
+while IFS= read -r f; do
```
manugarg, line 79: **Orphaned HTML directories pin bundles permanently — the "will drain" claim doesn't hold for `docs/config/master/`.**

The script only ever considers fingerprint-named files as candidates, so orphaned pages Hugo no longer emits keep their bundles alive forever. Not hypothetical: `origin/gh-pages` contains `docs/config/master/` (superseded by `docs/config/main/` when the branch was renamed), and e.g. `docs/config/master/alerting/index.html` references `index.min.3b3789d8….js` and `main.min.162c56a0….js`.

Nothing regenerates `master` — the `tag: all` dispatch in `latest_config_docs.yml:63` enumerates `docs/public/docs/config/v*.*.*` plus `git tag --list 'v*.*.*'`, which never yields `master` — so those bundles can never drain, contrary to the PR description. Worth either deleting `docs/config/master/` from gh-pages separately or noting the residue.

tools/clean_stale_assets.sh
```diff
@@ -0,0 +1,61 @@
...
+  base=${f##*/}
+  if ! grep -qxF "$base" <<<"$referenced"; then
+    if [[ "$DRY_RUN" == 1 ]]; then
+      echo "would delete: $f"
+    else
+      freed=$((freed + $(stat -c%s "$f")))
```
manugarg: **`stat` failure aborts mid-run and leaves the tree half-cleaned.**

`stat -c` is GNU-only. On macOS (the script is documented as locally runnable, and `-n` is offered as a dev flag but doesn't cover this path) `stat` errors, the arithmetic expansion becomes `$((freed + ))`, which is a syntax error, and `set -e` kills the script after some files have already been `rm`'d. Same outcome if the file disappears between `find` and `stat`.

`$(stat -c%s "$f" || echo 0)` or `wc -c` would be safer.

tools/clean_stale_assets.sh
```diff
@@ -0,0 +1,61 @@
...
+#
+# Only fingerprint-named files are ever considered, and only when
+# unreferenced, so files that live on in the tree without being rebuilt by
+# this run -- in particular the versioned config docs, which reference the
+# bundles they were built with -- keep working.
+
+set -euo pipefail
+
+DRY_RUN=0
+if [[ "${1:-}" == "-n" ]]; then
```
manugarg: **`-n` is only honoured as the first argument** — `clean_stale_assets.sh docs/public -n` deletes for real.

The flag check looks only at `$1`, and `$2` is ignored entirely with no error. For a destructive script, someone typing the flag in the more natural trailing position gets a live deletion of ~171 files instead of the dry run they asked for. Either parse flags in a loop or reject extra arguments.

manugarg (commented):
Code review of the stale-asset cleanup. I ran the script against the real `docs/public` worktree: the dry run finds 20 candidates and none of them is referenced by any surviving page, including pages from an older build that pin a different `index.min.<sha>.js`. The core matching logic looks right — the regex handles minified unquoted attributes and absolute `https://cloudprober.org/...` URLs, `sed 's|.*/||'` reduces to basenames, the loop uses process substitution so the counters survive, and `find -path ./.git -prune` is correct for a worktree. Four findings inline; the first is the one I'd fix before merging.

Also checked and cleared: no `.svg`, `.txt`, `.webmanifest`, `_headers`, or `_redirects` file references a fingerprinted asset, so the `--include` list is sufficient today; no scanned file is binary-detected by grep, so no references are silently dropped; Hugo 0.101 image variant specs contain only `[A-Za-z0-9_]`, so the params class is adequate; and both workflow steps run from the repo root, land between `Build pages` and `Deploy`, with `git add --all` staging the deletions.

tools/clean_stale_assets.sh
manugarg: Fixed in 26a940b0. The scan no longer swallows failures: grep exit 2 (read error, broken pattern) aborts the run, and zero references alongside existing candidates is treated as a broken tree/scan and aborts with an explicit error instead of feeding an empty set to the deletion loop. Verified your repro (a directory holding only the two fingerprinted files, no HTML): it now exits 1 and deletes nothing; a genuinely empty tree still no-ops with exit 0.

tools/clean_stale_assets.sh
manugarg: Agreed — done both ways. (1) Deleted docs/config/master/ from gh-pages as a one-off (8230d999) after confirming it is a pure fossil: zero inbound links from anywhere, not in the sitemap, and the version selector (config-docs-nav.html) only enumerates v*.*.* dirs, so nothing could link to it or regenerate it. The two bundles it exclusively pinned (index.min.3b37…, main.fc64…) are now collectable — the cleanup dry-run on current gh-pages flags them. (2) Corrected the PR description: draining only applies to pages that actually get rebuilt. Also noted there for follow-up: other unreachable top-level sections (concepts/, getting-started/, surfacers/, how-to/, goto/) — several are still link targets of current pages via stale URLs, so they need a content fix before removal.

tools/clean_stale_assets.sh
manugarg: Fixed in 26a940b0: size is now computed with wc -c < "$f" (portable, no GNU stat) and tolerates the file disappearing between find and read — the freed-bytes counter degrades to 0 instead of a syntax error killing the run mid-cleanup.

tools/clean_stale_assets.sh
manugarg: Fixed in 26a940b0: flags are parsed in a loop now. -n/--dry-run is honored in any position; unknown flags and extra positional arguments are rejected with a usage message. Verified: dir -n dry-runs, dir extra and -x dir both error out.

PR #1468, opened 2026-09-06
[docs] Track the logo kit and make it reproducible

manugarg:
Until now the logo existed in the repo only as rendered output under `docs/static`.
The geometry that generates it lived outside version control, which is why the icon
fix in #1466 had to be made by hand-editing a rendered SVG rather than changing a
parameter and rebuilding.

## Layout

| path | contents |
| ---- | -------- |
| `docs/brand/` | the artwork — 17 SVG variants, PNG renders, the favicon set, `PREVIEW.png`, and the usage rules in `README.md` |
| `tools/logogen/` | `build_logo.py` and the weight-600 IBM Plex Sans instance it builds from |

A generator under `tools/` writing into `docs/` matches what `tools/gen_config_docs.sh`
already does, and the `tools/logogen/` subdirectory follows `tools/genprotofacade/`.

Named `docs/brand` rather than `docs/logo` to keep it distinct from `docs/static/logo`,
which holds the published copies the site actually serves. Hugo publishes only
`content/` and `static/`, so `docs/brand` is not picked up by the site build or the
gh-pages tree.

## The kit is reproducible

```sh
pip install fonttools cairosvg pillow numpy
python3 tools/logogen/build_logo.py --verify
```

Building into a clean directory reproduces **all 42 generated files in `docs/brand`
byte-for-byte**, every PNG and the ICO included. The two files that are not generated,
and are safe to hand-edit, are `README.md` and `PREVIEW.png`.

`--out` and `--font` default relative to the script rather than the caller's cwd, so that
command rebuilds `docs/brand` in place from anywhere in the tree.

## What --verify checks

It exits non-zero on failure. Confirmed working by reintroducing the #1466 defect —
setting `ICON_POST_TOP` back to `141` is caught three independent ways:

```
- ICON_POST_TOP is 141; 144 is the value that clears the dish at stroke 14
- pre-fix icon post survives in cloudprober-icon{,-ondark,-white,-mono}.svg
- icon post breaks the dish surface in 84 columns
```

That last one rasterises the dish and the mount separately and compares them column by
column, so it measures the defect rather than pattern-matching a known-bad constant. It
also checks that no SVG carries live text or a font dependency, that all four
`favicon.ico` planes are true renders at their own size rather than upsamples, and that
`apple-touch-icon.png` is 180×180 and opaque.

## Superseded scripts

The kit originally came with `mark.py` and `build_final.py`. Those were helpers only — no
`main`, no file writing, no rasterisation — and carried absolute `/home/claude` paths from
the sandbox they were written in, so they could not run anywhere else. `build_logo.py`
supersedes both and they are removed rather than left as a trap.

## One correction to #1466

That PR justified repacking `favicon.ico` by saying the kit ships a 16×16-only `.ico`.
That was measured on a superseded copy of the kit and is wrong: the kit's `.ico` has all
four 16/32/48/64 planes and is byte-identical to the repacked file. The shipped asset is
correct and unaffected — only the stated reason was wrong.

## Verification

- All 42 generated files reproduce byte-for-byte from a clean build.
- The artwork is byte-identical to the copies under `docs/static`, across all 14
  published assets.
- `--verify` passes on the committed tree and fails on a deliberately regressed one.
- No `/home/claude` paths remain anywhere in the tree.

IBM Plex Sans is SIL OFL 1.1; the wordmark is outlined, so the reserved-font-name clause
does not apply to the artwork and no output file carries a font dependency.


manugarg:
All five addressed, plus the `-O` nit. Pushed as 2300a4c2.

**1 — `__pycache__` (blocking).** Correct on every point, including the diagnosis. Untracked both `.pyc` files and added explicit `__pycache__/` and `*.pyc` rules at the end of `.gitignore`, where they take precedence over `!*.*`. Confirmed with `git check-ignore`, and no bytecode remains tracked anywhere in the repo.

For the record on how they got in: they're bytecode from an import check I ran against `build_final.py` while it still existed, swept in by `git add tools/logogen`, and the later deletion of the sources didn't carry them out. So the branch was shipping compiled copies of the exact files it claimed to have removed, with absolute paths embedded, against its own "no absolute paths" claim.

**2 — `--instance-from`.** Real, and worth noting I introduced it: anchoring `--font` to the script directory is what made the default point *at* the tracked font. Before that it landed in the caller's cwd, which was merely useless rather than destructive. It now refuses unless `--font` names an explicit target:

```
$ python3 tools/logogen/build_logo.py --instance-from Variable.ttf
--instance-from needs an explicit --font to write to; it would otherwise
overwrite the tracked tools/logogen/IBMPlexSans-600.ttf.
Pass e.g. --font /tmp/IBMPlexSans-600.ttf, and copy it into place yourself
once you are happy with it.
```

Exits 1, and I checked the tracked font's hash is unchanged afterwards. An explicit `--font` still proceeds.

**3 — `/tmp` scratch files.** Removed rather than relocated: `cairosvg.svg2png` returns PNG bytes when `write_to` is omitted, so these go through `io.BytesIO` and no scratch file is created. That covers both the Windows path and the concurrency clobber without a tempdir to clean up.

**4 — `verify()` docstring.** Rewritten to describe what it does. It does read only from disk, so it would work on any tree laid out like the build output; the docstring now says that while making clear `--verify` rebuilds first.

**5 — Count.** You're right, 14. My 13 came from an earlier check that omitted `favicon.ico`. Fixed in the PR body; all 14 verified byte-identical.

**Nit — `assert` under `-O`.** Cheap enough to just fix: now an explicit `raise`. Confirmed the build passes under `python -O`.

**Nit — trailing newlines.** Agreed, leaving as is.

Re-verified after all of it: clean build reproduces all 42 generated files byte-for-byte, `--verify` passes, and reintroducing `ICON_POST_TOP=141` is still caught six ways including the 84-column dish measurement, exiting non-zero.

PR #1481, opened 2026-09-11
[logger] Never block the caller on writes to stderr

manugarg:
If whatever reads the other end of stderr stalls (for example, the container runtime's log writer on a node with a stuck disk), a write to stderr blocks indefinitely. Every goroutine that logs then blocks behind that one write, including probe goroutines, so probing gradually stops. We saw this in production: `cloudprober_total` stepped down to zero over about 9 minutes while the node's logs went dark.

### Change

- Loggers created without `WithWriter` now write to stderr through a new `asyncWriter`. A single goroutine writes the queued entries in order. `Write` copies its input and never blocks.
- If the queue fills up (4 MiB), new entries are dropped and counted. A WARN entry with `dropped_entries=N` goes just before the next accepted entry, so the gap is marked in the log stream.
- Critical logs, and `main`'s exit paths, flush the queue before exiting. The flush has a timeout, so a stalled stderr can't block the exit either.
- Writers passed with `WithWriter` stay synchronous.

This follows zerolog's `diode.Writer` (the input copy, drops reported in the stream) and Docker's non-blocking log mode (a byte budget, dropping the newest entries). It's a buffered queue with no new dependencies.

### Testing

- Unit tests for a blocked writer (writes don't block, drops are counted, order is preserved, the notice lands at the gap, `Flush` times out), for the input copy, and for concurrent writes. They pass 30 times in a row under `-race`, and the full test suite passes.
- Manual run: an HTTP probe with 20 failing targets at a 100ms interval, with stderr going to a FIFO whose reader never reads.
  - **Before:** probing froze within seconds, with no exports at all over 35s.
  - **After:** probing ran at full rate, with 100 → 200 → 300 per target across three 10s exports. Only the drain goroutine was blocked in `pipe_write`.

### Follow-ups

- Export the drop count as a sysvar metric. Metrics kept flowing during the incident while logs were dark, so this metric would have made the stall visible.
- Route the `raw_stderr_output` relays in the external and command probes through the same writer.

### DIFF
--- a/logger/asyncwriter.go
+++ b/logger/asyncwriter.go
@@ -24,8 +24,8 @@ import (
 	"time"
 )
 
-// maxQueuedLogBytes is the most log data we hold in memory while waiting for
-// stderr to accept it.
+// maxQueuedLogBytes (4 MiB) is the most log data we hold in memory while
+// waiting for stderr to accept it.
 const maxQueuedLogBytes = 4 << 20
 
 // asyncWriter is an io.Writer that never blocks the caller. Writes are queued
@@ -81,6 +81,9 @@ func (aw *asyncWriter) Write(p []byte) (int, error) {
 	aw.queued += len(p)
 	aw.mu.Unlock()
 
+	// Wake up the drain goroutine without blocking. If a wake-up is already
+	// pending, the send is skipped: the drain goroutine takes the whole queue,
+	// including this entry, when it handles that wake-up.
 	select {
 	case aw.ready <- struct{}{}:
 	default:
@@ -96,6 +99,8 @@ func (aw *asyncWriter) drain() {
 		aw.mu.Unlock()
 
 		for _, b := range batch {
+			// Nowhere to report a failed write to stderr; the synchronous
+			// path ignored these errors too.
 			aw.w.Write(b)
 			aw.mu.Lock()
 			aw.queued -= len(b)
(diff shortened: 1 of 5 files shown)


logger/asyncwriter.go
```diff
@@ -0,0 +1,132 @@
...
+	"io"
+	"log/slog"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// maxQueuedLogBytes is the most log data we hold in memory while waiting for
+// stderr to accept it.
+const maxQueuedLogBytes = 4 << 20
```
manugarg: In comment, add what 4 << 20 resolves to (4MB).

logger/asyncwriter.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	// Copy p: slog handlers reuse their buffers after Write returns.
+	aw.queue = append(aw.queue, bytes.Clone(p))
+	aw.queued += len(p)
+	aw.mu.Unlock()
+
+	select {
+	case aw.ready <- struct{}{}:
+	default:
+	}
```
manugarg, line 90: Add a comment to this section. IIUIC, this is where the non-blocking part comes to picture. We nudge the drainer but don't block on it?

logger/asyncwriter.go
```diff
@@ -0,0 +1,132 @@
...
+	var buf bytes.Buffer
+	r := slog.NewRecord(time.Now(), slog.LevelWarn, "Dropped log entries because writes to stderr were blocked", 0)
+	r.AddAttrs(slog.String("system", defaultSystemName), slog.Int64("dropped_entries", n))
+	slogHandler(&buf).Handle(context.Background(), r)
+	return buf.Bytes()
+}
+
+// Flush waits until all queued entries have been written, or until timeout
+// expires. It reports whether the queue was fully written.
+func (aw *asyncWriter) Flush(timeout time.Duration) bool {
```
manugarg: Given that Flush doesn't call drain itself, this function name is a bit confusing. How about Wait() or something else?

logger/asyncwriter.go
```diff
@@ -0,0 +1,132 @@
...
+		queued := aw.queued
+		aw.mu.Unlock()
+
+		if queued == 0 {
+			return true
+		}
+		if time.Now().After(deadline) {
+			return false
+		}
+		time.Sleep(10 * time.Millisecond)
```
manugarg, line 135: What if we are already after the deadline before the loop even starts? We create our own deadline, but I'd prefer if we do this check before the loop start, wdyt?

Also maybe implementing. isQueuedZero() make it more idiomatic. I am not sure.

logger/asyncwriter.go
manugarg: Fixed.

logger/asyncwriter.go
manugarg: Done.

logger/asyncwriter.go
manugarg: Done.

logger/asyncwriter.go
manugarg: Done.

PR #1484, opened 2026-09-12
[rds/kubernetes] Time out API server calls, and share one lister across the four resource types

manugarg:
## Timeout (first commit)

The kubeapi client was built without an `http.Client` timeout and its requests carry no context, so only the cloned `DefaultTransport`'s dial (30s) and TLS handshake (10s) timeouts applied. An API server that completes the handshake and then never sends response headers hung `getURL` forever.

Each lister refreshes from a single goroutine driven by a ticker, so one stuck call froze that resource type's refresh permanently — targets went stale silently and never recovered without a restart.

Sets a 30s overall timeout, comfortably under the 60s default `re_eval_sec`. `Client.Timeout` covers the whole request lifecycle including the body read in `getURL`, and survives `newClient` wrapping the transport in `oauth2.Transport`.

## Deduplication (second commit)

pods, endpoints, services and ingresses each carried their own copy of the same lister: an identical struct, a URL builder differing only in the API prefix and resource kind, a `parseXJSON`, an `expand()`, and a byte-identical refresh goroutine.

All four are now a generic `resourceLister[T resourceInfo]` in `lister.go`. Each resource type supplies only what actually differs: its metadata, and how it expands into RDS resources. Net -392 lines.

### Cache wipe on error

The four copies of `expand()` shared a bug: on an API or parse error they logged a warning and fell through to overwrite the cache with `nil`, so a transient failure dropped every target for that resource type until the next successful refresh.

There's no net for this downstream — an emptied cache reaches the RDS client as a *successful* response with zero resources, not as an error, so `updateState` faithfully replaces its cache with empty.

This mattered little while calls could hang, since a hang preserved the cache. The timeout above turns API stalls into exactly the kind of fast error that hits this path, so `expand()` now keeps the last good result.

### Preserved asymmetries

- **Ingresses** filter name/labels *after* expansion, not on the Kubernetes object, because their resource names are derived per rule (`<ingress>_<host>_<path>`) and they carry their own `fqdn`/`relative_url` labels. The other types filter on the object, where a name filter matches the object name rather than derived per-port resource names. Flattening this either way would silently change target selection.
- **Pods** ignore a name in the resource path (`pods/pod-a`), which the other three honor. Kept behind the `nameInPath` field so existing configs don't silently narrow from all pods to one. Happy to unify this if you'd prefer — it's a one-word change.

### Also

Drops `rand.Seed` (deprecated since Go 1.20, called once per lister at startup) and jitters the first refresh at full resolution instead of whole seconds; drops the unused per-type config fields, as all four config messages are empty; drops the `error` return from the four constructors, which was never non-nil.

## Testing

The listers' test assertions are unchanged — only the call sites of the removed helpers were updated — which is the evidence the refactor preserves behavior. Added `TestGetURLTimeout` (httptest TLS server that blocks until the request context is cancelled) and an assertion that the constructed client carries the timeout.

`go build ./...`, `go test ./internal/rds/...` and `go test -race ./internal/rds/kubernetes/` all pass.

## Known gap, not addressed

`re_eval_sec: 0` panics in a refresh goroutine via `time.NewTicker(0)` and takes the process down. It panicked before this change too (`rand.Intn(0)`), so it's not a regression, but it's now in one place and would be cheap to turn into a config error out of `New()`.

### DIFF
--- a/internal/rds/kubernetes/client.go
+++ b/internal/rds/kubernetes/client.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/cloudprober/cloudprober/common/oauth"
 	"github.com/cloudprober/cloudprober/common/tlsconfig"
@@ -36,6 +37,11 @@ var (
 	LocalCACert    = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 	LocalTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 )
+
+// apiCallTimeout is the overall timeout (connection, headers, and body) for a
+// call to the Kubernetes API server. Without it, an API server that accepts
+// the connection but never responds stalls the lister's refresh loop forever.
+const apiCallTimeout = 30 * time.Second
 
 // client encapsulates an in-cluster kubeapi client.
 type client struct {
@@ -143,6 +149,7 @@ func newClientWithoutToken(cfg *configpb.ProviderConfig, l *logger.Logger) (*cli
 
 	c.httpC = &http.Client{
 		Transport: transport,
+		Timeout:   apiCallTimeout,
 	}
 
 	if err := c.initAPIHost(); err != nil {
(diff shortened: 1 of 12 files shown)


manugarg:
Split into two independent PRs so each can be iterated on separately: #1485 (API server timeout) and #1486 (shared generic lister). They touch no overlapping files and can merge in either order.

PR #1486, opened 2026-09-12
[rds/kubernetes] Share one generic lister across the four resource types

manugarg:
pods, endpoints, services and ingresses each carried their own copy of the same lister: an identical struct, a URL builder differing only in the API prefix and resource kind, a `parseXJSON`, an `expand()`, and a byte-identical refresh goroutine.

All four are now a generic `resourceLister[T resourceInfo]` in `lister.go`. Each resource type supplies only what actually differs — its metadata, and how it expands into RDS resources:

```go
type resourceInfo interface {
	metadata() kMetadata
	resources(f *listFilters, l *logger.Logger) []*pb.Resource
}
```

## Cache wipe on error

The four copies of `expand()` shared a bug: on an API or parse error they logged a warning and fell through to overwrite the cache with `nil`, dropping every target for that resource type until the next successful refresh.

There's no net for this downstream — an emptied cache reaches the RDS client as a *successful* response with zero resources, not as an error, so `updateState` faithfully replaces its cache with empty.

It mattered little while calls could hang, since a hang preserved the cache. It matters more once API calls time out (#1485), because a stall then surfaces as exactly this kind of fast error. `expand()` now keeps the last good result.

That PR and this one touch no overlapping files and can merge in either order.

## Behavior changes

- **A name in the resource path now selects a single pod.** `k8s://pods/my-pod` used to return every pod in the cluster; services, endpoints and ingresses have always treated the name as a selector. Pods now match them. A config using such a path gets the one pod it asked for rather than all of them. This is the only change here that can alter an existing deployment's target set.
- **`re_eval_sec` must be positive.** An explicit `0` used to panic the refresh goroutine on `time.NewTicker(0)`, taking the process down; it is now a config error from `New()`. An unset value still defaults to 60, and `k8s_targets` forces 30, so only an explicit zero or a negative value is affected.

## Preserved asymmetry

Ingresses filter name/labels *after* expansion, not on the Kubernetes object, because their resource names are derived per rule (`<ingress>_<host>_<path>`) and they carry their own `fqdn`/`relative_url` labels. The other types filter on the object, where a name filter matches the object name rather than derived per-port resource names. Flattening this either way would silently change target selection.

## Also

Drops `rand.Seed` (deprecated since Go 1.20, called once per lister at startup) and jitters the first refresh at full resolution instead of whole seconds; drops the unused per-type config fields, as all four config messages are empty; drops the `error` return from the four constructors, which was never non-nil; documents the `keys`/`cache` invariant that `listResources` relies on when it dereferences a cache entry.

## Testing

Coverage went from 65.2% to 83.9%. The starting number was badly distributed: the existing tests covered the filtering and expansion logic well, and everything the refactor *introduced* — `url()`, `expand()`, `start()`, all four constructors — was at zero, so a wrong api prefix or resource kind would compile, pass every test, and silently return no targets.

The listers' original test assertions are unchanged, which is the evidence the refactor preserves behavior. New tests cover the API path each lister requests (with and without a namespace), `expand()` keeping the last good result on API and parse failures, the resource-path selector, the running-pods filter, and the `re_eval_sec` rejection — all driven through the real constructors so the per-type wiring is observable.

Verified by mutation rather than by the coverage number: breaking the implementation eleven ways (wrong api prefix, wrong kind, `url()` ignoring the namespace, `expand()` reverted to the old fall-through, `keep` removed, `runningPod` accepting every phase, `matches()` ignoring either filter, `listResources` dropping the namespace filter or the path selector, ingresses filtering the object instead of expanded resources) fails the tests in every case.

Still uncovered, pre-existing and not addressed: `ListResources()` in `kubernetes.go`.

## Review

A cold review of the branch raised three items, all addressed: the half-guarded `NewTicker` panic (now rejected in `New()`), the pods resource-path inconsistency (now unified), and a stale doc reference to `listFilters.matchesObject` left over from collapsing two identical match helpers into `matches`.

### DIFF
--- a/internal/rds/kubernetes/endpoints.go
+++ b/internal/rds/kubernetes/endpoints.go
@@ -103,12 +103,11 @@ func (epi *epInfo) resources(f *listFilters, l *logger.Logger) (resources []*pb.
 
 func newEndpointsLister(namespace string, reEvalInterval time.Duration, kc *client, l *logger.Logger) *epLister {
 	lister := &epLister{
-		kind:       "endpoints",
-		apiPrefix:  "api/v1",
-		namespace:  namespace,
-		kClient:    kc,
-		nameInPath: true,
-		l:          l,
+		kind:      "endpoints",
+		apiPrefix: "api/v1",
+		namespace: namespace,
+		kClient:   kc,
+		l:         l,
 	}
 	lister.start(reEvalInterval)
 	return lister
--- a/internal/rds/kubernetes/endpoints.go
+++ b/internal/rds/kubernetes/endpoints.go
@@ -15,81 +15,16 @@
 package kubernetes
 
 import (
-	"encoding/json"
 	"fmt"
-	"math/rand"
 	"strconv"
-	"strings"
-	"sync"
 	"time"
 
-	configpb "github.com/cloudprober/cloudprober/internal/rds/kubernetes/proto"
 	pb "github.com/cloudprober/cloudprober/internal/rds/proto"
-	"github.com/cloudprober/cloudprober/internal/rds/server/filter"
 	"github.com/cloudprober/cloudprober/logger"
 	"google.golang.org/protobuf/proto"
 )
 
-type epLister struct {
-	c         *configpb.Endpoints
-	namespace string
-	kClient   *client
-
-	mu    sync.RWMutex // Mutex for names and cache
-	keys  []resourceKey
-	cache map[resourceKey]*epInfo
-	l     *logger.Logger
-}
-
-func epURL(ns string) string {
-	if ns == "" {
-		return "api/v1/endpoints"
-	}
-	return fmt.Sprintf("api/v1/namespaces/%s/endpoints", ns)
-}
-
-func (lister *epLister) listResources(req *pb.ListResourcesRequest) ([]*pb.Resource, error) {
-	var resources []*pb.Resource
-
-	var epName string
-	tok := strings.SplitN(req.GetResourcePath(), "/", 2)
-	if len(tok) == 2 {
-		epName = tok[1]
-	}
-
-	allFilters, err := filter.ParseFilters(req.GetFilter(), SupportedFilters.RegexFilterKeys, "")
-	if err != nil {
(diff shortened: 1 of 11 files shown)


manugarg:
Added a second commit with the coverage this refactor was missing.

**Before:** 65.2% package coverage, but badly distributed. The existing tests pinned the filtering and expansion logic well (`resources()`, `matches()`, `listResources`, `parseResourceList` were all 90%+) — and everything the refactor actually *introduced* was at zero: `url()`, `expand()`, `start()`, and all four constructors. A wrong `apiPrefix`, `kind`, `nameInPath` or `keep` would compile, pass every existing test, and silently return no targets in production.

**After:** 82.6%. Everything in `lister.go` is covered except `start()`'s ticker loop.

New tests: the API path each lister requests (with and without a namespace), `expand()` keeping the last good result on API and parse failures, pods ignoring a name in the resource path while services honor it, and the pods lister caching only running pods. All of them go through the real constructors rather than hand-built structs, which is what makes the per-type wiring observable.

**Verified by mutation** rather than by the coverage number — I broke the implementation eight ways and confirmed each one fails:

| mutation | caught |
|---|---|
| ingresses `apiPrefix` → `api/v1` | yes |
| endpoints `kind` → `pods` | yes |
| `url()` ignores namespace | yes |
| `expand()` reverted to the old fall-through | yes |
| pods `keep` filter removed | yes |
| `runningPod` accepts every phase | yes |
| `matches()` ignores the name filter | yes |
| `matches()` ignores the labels filter | yes |
| `listResources` drops the namespace filter | yes |
| ingresses filter the object instead of expanded resources | yes |

One note for review: in `expand()`, the early `return` after the `getURL` error is not independently observable, because `getURL` always returns a nil body alongside its error, so the parse-error `return` backstops it. Removing *either* one alone still passes; removing both reproduces the original bug and fails. Keeping both as defense in depth — worth knowing that the first one isn't load-bearing on its own.

Still uncovered and pre-existing, not addressed here: `New()` and `ListResources()` in `kubernetes.go`.

PR #1490, opened 2026-09-13
[servers.udp] Support large-datagram in echo server

muff1nman:
Allow configuring the maximum udp payload that can be received.

### DIFF
--- a/internal/servers/udp/udp.go
+++ b/internal/servers/udp/udp.go
@@ -39,19 +39,14 @@ const (
 
 	// Number of messages to batch read.
 	batchSize = 16
-
-	// Maximum packet size.
-	// TODO(manugarg): We read and echo back only 4098 bytes. We should look at raising this
-	// limit or making it configurable. Also of note, ReadFromUDP reads a single UDP datagram
-	// (up to the max size of 64K-sizeof(UDPHdr)) and discards the rest.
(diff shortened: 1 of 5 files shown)


internal/servers/udp/udp.go
```diff
@@ -90,6 +85,11 @@ func (s *Server) configureAdvancedReadWrite() error {
 
 // New returns an UDP server.
 func New(initCtx context.Context, c *configpb.ServerConf, l *logger.Logger) (*Server, error) {
+	maxPayloadSize := c.GetMaxPayloadSize()
+	if maxPayloadSize < 1 || maxPayloadSize > 65535 {
```
manugarg, line 89: Nit: the actual maximum UDP payload is 65507 bytes over IPv4 (65535 total - 20-byte IP header - 8-byte UDP header; 65527 for IPv6), so values above that can never be observed on the wire. The 65535 bound is harmless (it just allows a slightly oversized buffer allocation), but we could tighten the validation and the proto comment/docs to 65507 if you prefer.

muff1nman: Right I did consider doing that but figured might as well support the full udp packet size in case someone comes up with some esoteric way to deliver packets to this application.

muff1nman:
@manugarg are you good to proceed with this?

PR #1494, opened 2026-09-22
[tracing] Flush spans on --run_once, and bound the flush

manugarg:
Follow-up to #1358.

### `--run_once` never flushed spans

`tracing.Init` returns `tp.Shutdown` for a batching provider, and the only
caller was the `ctx.Done()` goroutine in `Start()`. `--run_once` returns from
`main` without ever calling `Start`, so every span from a single run was
dropped.

This adds an exported `cloudprober.Shutdown()` and calls it from main's
`run_once` branch. It's an explicit call *before* the error check, not a
`defer` after `Init()`: `l.Criticalf` exits the process without running
defers, and a failing probe is exactly the case where you want its spans.

### Bounding the flush

All shutdown call sites go through a shared `shutdownTracing()` helper with a
3s timeout. The init rollback path previously used `context.Background()`
while holding the write lock, so an unreachable collector (which the exporter
retries with a backoff) could hold up initialization. And 3s keeps the flush
below the default `stop_time_sec` of 5s: at 5s the flush deadline and main's
`os.Exit(0)` expired at the same instant, so the "Error shutting down tracing"
warning was unreachable in practice.

### `Shutdown()` is the one teardown path, and it's one-way

`Start()`'s `ctx.Done()` goroutine used to do its own cleanup, overlapping
with what `Shutdown()` needed to do. The whole teardown now lives in
`Shutdown()` -- default servers, probes, init context, tracing flush -- and
the goroutine is just `<-ctx.Done(); Shutdown()`. `Start()` runs the prober on
a context derived from the caller's, so `Shutdown()` can stop the probes even
when the caller's context is still live.

The RunOnce path consequently releases the default HTTP/gRPC listeners that
`Init` opens for it, which embedders calling `Init` + `RunOnce` had no way to
do before.

Everything `Init` acquires lives in one `instance` struct, so a shutdown swaps
in a zero value rather than clearing fields one at a time.

### Behavior change: cloudprober can be initialized only once

`Shutdown()` is one-way, and `Init` now **panics** if it's called a second
time, whether or not a shutdown has happened (it used to silently return
`nil` when already initialized). Cloudprober is a process-wide singleton and
we don't restart it; config reload will reload individual components rather
than the whole prober.

This drops `TestRestart`, added in a905e2c4 (2021) so embedders could stand up
a full cloudprober per test case. What it actually verified -- that a teardown
gives the ports back -- is now covered by `TestShutdownWithoutStart` and
`TestShutdownStopsStartedProber`.

`RunOnce` also returns an error instead of panicking when the prober isn't
initialized, which covers calling it before `Init` as well.

### Verification

Against a fake OTLP/HTTP collector:

- `--run_once`, pre-change binary: no export at all; with the change: spans
  exported, on both the success path and the probe-failure path that ends in
  `Criticalf`.
- the `Start` path still flushes on SIGTERM.
- black-holed collector (`240.0.0.1:4318`): run exits in 3.1s with
  `WARN Error shutting down tracing err="context deadline exceeded"`.

Tests cover the flush (nil, once-only, deadline, error), teardown without
`Start`, teardown of a started prober including that its ports come back, and
the init-after-shutdown panic. `go test -race` is clean.

### Not in this change

A `stop_time_sec: 0` config installs no signal handler in `main` at all, so
SIGTERM kills the process with spans unflushed. That's a separate fix.


cbroglie:
LGTM 👍

